PR #2, opened 2015-09-21
Document windows usage

msarahan:


faph:
I was just looking through this PR because I was interested to see how `menuinst` actually works. I think the build script will need to copy the menu.json to a "Menu" folder in %PREFIX%. That way `conda install` will pick it up when installing. I'm using this and that seems to work:

```cmd
python setup.py install
mkdir %PREFIX%\Menu
copy %RECIPE_DIR%\menu-windows.json %PREFIX%\Menu
if errorlevel 1 exit 1
```
Also, does the menu.json file not need to have the full structure menuinst is expecting, something like this?

```json
{
    "menu_name": "World",
    "menu_items":
        [
            {
                "name": "Hello World",
                "pyscript": "${PYTHON_SCRIPTS}/helloworld-script.py"
            }
        ]
}
```

msarahan:
Are you sure copying the menu.json file is necessary?  I think it is sufficient for the menu-windows json to simply be part of the recipe folder, and it will get found.  I might be wrong though - this is probably a subtle difference between Conda and our internal build system, which predated Conda.  Copying it seems like the best compromise between the two; thanks for pointing this out.

Regarding the menu.json structure, I think what you're asking is "is menu_name and menu_items required?"  Is that right?  If so, the answer is no.  When they are not provided, they default to the Anaconda menu.  I will add a bit about this - thanks for pointing it out.

faph:
Yes, I'm just doing plain `conda build` and `conda installs`, so that might be different. Searching through the conda build source code i don't see anything where menu items are parsed. It doesn't matter I think because it can perfectly be handled by the build scripts.

faph:
It's a shame that `menuinst` is not well documented. It seems a good library. If all conda apps provide this we don't really need the launcher, apart from creating envs and installing packages.

msarahan:
I think it works well for Windows, but perhaps less so for Mac/Linux?  Maybe the maintenance across platforms is easier with the anaconda launcher?  This is speculation on my part, but I agree with you that we can do a better job documenting menuinst for those who want to use it.

faph:
I've just started scanning through the code and collecting the menu+shortcut config spec, see: https://github.com/ContinuumIO/menuinst/wiki/Menu-Shortcut-Config-Structure if that's ok.

msarahan:
Excellent!  I think that's a nice addition.  Perhaps better to keep it
there, and make a link to that page in the README

faph:
I've now also added a page with conda build script examples. I'm on Windows, so have only been able to actually test the `bld.bat`. I've created a test package at https://anaconda.org/faph/helloworld and at https://github.com/faph/Hello-World. 

As far as I can see it, for linux menu items you cannot configure shortcuts to scripts in the `$PREFIX` folder as you don't have any variable you can use. For Windows you have `${PREFIX}` and for OSX you have `${BIN_DIR}` amongst other variables.

PR #4, opened 2015-10-12
add suffix for environment name to shortcuts on win

msarahan:
NOTE: this might have issues with CONDA_DEFAULT_ENVs that are full paths, rather than short names.  Probably need to explicitly split off the last string following a \

See PR https://github.com/conda/conda-env/pull/60 which makes all CONDA_DEFAULT_ENV set to full paths.


faph:
Bear in mind that at the moment `menuinst` is fully independent of conda. This way you'll tie it to conda which may not be the preferred approach. Alternatively conda-build could handle any conda-specific issues such as manipulating menu titles or shortcut names. Another alternative would be to have conda-install handle this. Personally I think it would be nice to keep menuinst independent of conda.

msarahan:
Good point.  I'll see about rearranging things appropriately.  Menuinst might still need modification to support extra info (perhaps from virtualenv?), but it should not be specific to conda.

msarahan:
This is part of a 4-part set of PRs to allow shortcuts to be created from environments.  Other PRs:

  * https://github.com/conda/conda/pull/1699
  * https://github.com/conda/conda-env/pull/178

(more coming as they are added)

PR #15, opened 2015-10-30
rework shortcut env setting to use python instead of bat files

msarahan:


msarahan:
ping @ilanschnell @groutr

PR #24, opened 2016-01-06
delay folder deletion on Win to avoid race condition

msarahan:
ping @ilanschnell @csoja 

This is a vile hack.  I'm sorry.  I don't know what else to do here.  The issue is that people are not getting menus installed sometimes.  If you enable menuinst.DEBUG, then it always works.  If you comment out the shutil.rmtree, then it always works.  I'm pretty sure this is some awful Windows asynchronous file I/O race condition.  subprocess.check_call **should** absolutely be synchronous, and running the Python should be as well.  There should be no race condition here, but everything I see points to one.  If you have better ideas of how to fix this, I'd be excited to learn a better way.

### DIFF
--- a/menuinst/__init__.py
+++ b/menuinst/__init__.py
@@ -8,6 +8,7 @@ import json
 import shutil
 import subprocess
 import tempfile
+import time
 from os.path import abspath, basename, exists, join
 
 from ._version import get_versions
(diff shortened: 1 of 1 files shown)


ilanschnell:
Why not just keep the temporary directory around?  The directory is created using mkdtemp, and only contains two tiny files.  Not cleaning up the directory won't hurt anyone.

msarahan:
Cool, I'm very much OK with that solution, and that's much better than an arbitrary delay.

ilanschnell:
Done
https://github.com/ContinuumIO/menuinst/commit/2f37e3e24845d479dc73eb2e6a9b93c23d402c3b

PR #26, opened 2016-03-17
Add BSD 3-clause license as an explicit file

scw:
Having an explicit license document helps auditing efforts as folks won't always know to look into the setup.py file for licensing details.


msarahan:
Good call, thanks.

PR #27, opened 2016-04-07
Rework elevation - no more bat file; add KNOWNFOLDERS support

msarahan:
Our old CSIDL code is deprecated (https://msdn.microsoft.com/en-us/library/windows/desktop/bb762494(v=vs.85).aspx), and in at least one instance from the Anaconda Support mailing list, a CSIDL path was not found.  The new way is called KNOWNFOLDERID (https://msdn.microsoft.com/en-us/library/windows/desktop/dd378457(v=vs.85).aspx), and this PR brings it to Menuinst.  Win XP uses CSIDL, everything beyond that uses KNOWNFOLDERID.

This rips out the intermediate bat file, in favor of a more Python-based solution.  Doing so introduces a dependency on pywin32.

Finally, high-level integration tests are performed - installing shortcuts as user and as admin.  You should see interactive elevation notices when running the tests on platforms above XP.  If you don't, you might have UAC disabled - please enable it for these tests to make sure that it works.

This PR now addresses unicode support.  I have tested this by creating a user named "å test user".  When logged on as this user, the new tests here run as both user and admin.  I have reasonably high confidence that this fixes the issues with menu installation.  _nsis.py will need fixing (in both constructor and in the anaconda build system installer), also, though, since modifying PATH often fails when menu installation fails.

ping @ilanschnell @csoja


mingwandroid:
I can check it when I get back on Thursday.

mingwandroid:
Should be good to go now.

PR #29, opened 2016-05-24
Rdonnelly/sccm menuinst

mingwandroid:
@msarahan, @ilanschnell please review. This commit is the menuinst side of https://github.com/ContinuumIO/anaconda/pull/511

It fixes SCCM installs when done as SYSTEM user, which isn't something we should recommend since only Administrators can then update conda, however there are valid use-cases for this and it's better to succeed than to except.

I will look into sorting out some documentation regarding this later in the week.


msarahan:
LGTM.  Nice work refactoring the prefix and env name to live on the menu, rather than the shortcut.  That is a better way to do it.

PR #33, opened 2016-05-26
win32: Disable QuickLaunch for SYSTEM install

mingwandroid:
@msarahan, @ilanschnell, this part got lost in the last commit.

The QuickLaunch folder id doesn't exist for the
'nt authority/system' account and querying it
causes an exception, therefore we cannot add
QuickLaunch entries for this user.


msarahan:
LGTM.  Thanks @mingwandroid

PR #34, opened 2016-06-17
move knownfolder import out of cwp to fix import error

msarahan:
Menuinst was creating invalid shortcuts (perhaps failing to create them) due to import errors with knownfolders.  The problem was that the python running the shortcut is the one in the environment, which loses track of menuinst in PYTHONPATH.

This PR also adds in the MSYS2 paths for shortcuts, which we have been adding elsewhere.

@ilanschnell @mingwandroid - please review

@Maggie-M @csoja this might be what's causing the installers to fail with the menu errors.  The PATH errors are different, though. I'll discuss that elsewhere.

### DIFF
--- a/cwp.py
+++ b/cwp.py
@@ -6,24 +6,20 @@ import sys
 import subprocess
 from os.path import join
 
-from menuinst.knownfolders import FOLDERID, get_folder_path, PathNotFoundException
-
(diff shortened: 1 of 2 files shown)


mingwandroid:
Lgtm

msarahan:
This is no longer relevant.  Closing.

PR #36, opened 2016-10-04
cwp.py: Add library/mingw-w64/bin to PATH

mingwandroid:
This fixes https://github.com/ContinuumIO/anaconda-issues/issues/1129, although I would like to know why we cannot activate the root environment instead. IMHO, either activate/deactivate are used exclusively or they are gotten rid of entirely. Any middle ground is just bugs waiting to happen.


mingwandroid:
@ilanschnell, @msarahan, please review, I have packages ready to upload.

PR #39, opened 2017-01-10
fix fallback to user-mode shortcut if elevation denied

msarahan:
Replaces exceptions that look like

```
  File "<string>", line 1, in <module>
  File "c:\menuinst\menuinst\__init__.py", line 57, in install
    "import menuinst; menuinst.install(%r, %r, %r)" % (path, bool(remove), prefix)])
  File "c:\menuinst\menuinst\win_elevate.py", line 70, in runAsAdmin
    lpParameters=params)
pywintypes.error: (1223, 'ShellExecuteEx', 'The operation was canceled by the user.')
```

with log messages that look like

```
WARNING:root:Insufficient permissions to write menu folder.  Falling back to user location
```

HOWEVER: we don't keep a persistent record of where shortcuts are installed.  This PR handles the single-user with a system-wide install use case well.  It does not handle the multi-user system-wide install case.  In that case, each user can use this to create their shortcuts, but those shortcuts will not be removed if the sysadmin every uninstalls Anaconda/Miniconda.

### DIFF
--- a/menuinst/__init__.py
+++ b/menuinst/__init__.py
@@ -3,11 +3,11 @@
 # All rights reserved.
 
 from __future__ import absolute_import
+import logging
 import sys
 import json
-import subprocess
-import tempfile
 from os.path import abspath, basename, exists, join
+
 
 from ._version import get_versions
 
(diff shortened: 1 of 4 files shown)


ilanschnell:
LGTM.  Thanks!

PR #40, opened 2017-01-23
1.4.4 branch

mingwandroid:
@masrahan, @ilanschnell please review, we need this for RStudio to work better from Anaconda Navigator or when launched directly from the Start Menu shortcut.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,3 +1,8 @@
+2017-01-23: 1.4.4:
+------------------
+  * Set CONDA_PREFIX environment variable in cwp.py, fix PATH
+
+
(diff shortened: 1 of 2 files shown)


msarahan:
Looks fine to me.

ilanschnell:
I'm fine with it too (although I think it was more readable before).

mingwandroid:
Thanks guys. re: readability, did you check the individual patches to determine that or the net effect of the two combined? Because I added an experimental thing in the first commit then deleted it in the second one just so I'd have a record of it in-case I need it again. I hope that the sum difference isn't considered much (if any) less readable!

PR #47, opened 2017-05-30
fix non-root pythons being used to call cwp.py

### LINKED ISSUE #46: Incompatible cwp.py and menuinst package
Start menu shortcuts pull in multiple components to open a Python application, for example:

    C:\path\to\venv\pythonw.exe \                # the Python executable from the VIRTUAL env
    "C:\Program Files\Anaconda3\cwp.py" \        # the cwp script from the ROOT env
    C:\path\to\venv \                            # the VIRTUAL env directory
    "C:/path/to/venv/pythonw.exe" \              # the Python executable from the VIRTUAL env
    "C:/path/to/venv/Scripts/spyder-script.py"   # the application script from the VIRTUAL env

As you can easily see, the call of the `cwp.py` script is done through the interpreter from the virtual environment. What happened on my computers is that the `menuinst` package (which currently has to be installed in the root env) was not compatible to the `cwp.py` script which made the whole shortcut fail silently.

Shouldn't the first part call the Python interpreter from the ~~virtual~~ *root* environment instead of the interpreter from the ~~root~~ *virtual* environment? Then, the `menuinst` package from the root environment would be used.

msarahan:
fixes #46 

Thanks to @mbargull for figuring this out.

### DIFF
--- a/menuinst/win32.py
+++ b/menuinst/win32.py
@@ -7,7 +7,7 @@ from __future__ import absolute_import, unicode_literals
 import ctypes
 import logging
 import os
-from os.path import expanduser, isdir, join, exists, dirname
+from os.path import expanduser, isdir, join, exists
 import pywintypes
 import sys
 import locale
@@ -23,9 +23,11 @@ from .winshortcut import create_shortcut
 OutputDebugString = ctypes.windll.kernel32.OutputDebugStringW
 OutputDebugString.argtypes = [ctypes.c_wchar_p]
 
+
 class DbgViewHandler(logging.Handler):
     def emit(self, record):
         OutputDebugString(self.format(record))
+
 
 logger = logging.getLogger("menuinst_win32")
 logger.setLevel(logging.DEBUG)
@@ -87,7 +89,8 @@ def folder_path(preferred_mode, check_other_mode, key):
         if check_other_mode:
             logger.info("     .. despite 'check_other_mode'\n"
                         "        and 'other_mode' 'path' of '%s'\n"
-                        "        it excepted with: '%s' in knownfolders.py" % (path, type(exception).__name__))
+                        "        it excepted with: '%s' in knownfolders.py" % (path,
+                                                                    type(exception).__name__))
         else:
             logger.info("     .. 'check_other_mode' is False,\n"
                         "        and 'other_mode' 'path' is '%s'\n"
@@ -178,7 +181,7 @@ class Menu(object):
             used_mode = mode if mode else ('user' if exists(join(self.prefix, u'.nonadmin')) else 'system')
         else:
             used_mode = 'system'
-        logger.info("Menu: name: '%s', prefix: '%s', env_name: '%s', mode: '%s', used_mode: '%s'"
+        logger.debug("Menu: name: '%s', prefix: '%s', env_name: '%s', mode: '%s', used_mode: '%s'"
                     % (name, self.prefix, env_name, mode, used_mode))
         try:
             self.set_dir(name, self.prefix, env_name, used_mode)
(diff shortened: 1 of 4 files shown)


menuinst/win32.py
```diff
@@ -251,26 +254,28 @@ def remove(self):
...
+        # each of these roll the subprocess cmd into args
         if "pywscript" in self.shortcut:
-            cmd = join(self.menu.prefix, u"pythonw.exe").replace("\\", "/")
+            subprocess_cmd = join(self.menu.prefix, u"pythonw.exe").replace("\\", "/")
             args = self.shortcut["pywscript"].split()
-            args = get_python_args_for_subprocess(self.menu.prefix, args, cmd)
+            args = get_python_args_for_subprocess(self.menu.prefix, args, subprocess_cmd)
         elif "pyscript" in self.shortcut:
-            cmd = join(self.menu.prefix, u"python.exe").replace("\\", "/")
+            subprocess_cmd = join(self.menu.prefix, u"python.exe").replace("\\", "/")
```
mbargull: This is now a behavioral change in that previously `python.exe` instead of `pythonw.exe` called `cwp` -- though I don't if there are any implications in using `pythonw` instead..

menuinst/win32.py
```diff
@@ -251,26 +254,28 @@ def remove(self):
...
             args = self.shortcut["pyscript"].split()
-            args = get_python_args_for_subprocess(self.menu.prefix, args, cmd)
+            args = get_python_args_for_subprocess(self.menu.prefix, args, subprocess_cmd)
         elif "webbrowser" in self.shortcut:
-            cmd = join(unicode_prefix, u'pythonw.exe')
+            subprocess_cmd = join(unicode_prefix, u'pythonw.exe')
```
mbargull: This is now a no-op since `subprocess_cmd` isn't used afterwards.

menuinst/win32.py
```diff
@@ -251,26 +254,28 @@ def remove(self):
...
+            args.append(self.shortcut["script"].replace('/', '\\'))
             extend_script_args(args, self.shortcut)
-            args = get_python_args_for_subprocess(self.menu.prefix, args, cmd)
-            cmd = join(unicode_prefix, u"pythonw.exe").replace("\\", "/")
+            args = get_python_args_for_subprocess(self.menu.prefix, args, subprocess_cmd)
         elif "system" in self.shortcut:
-            cmd = substitute_env_variables(
-                     self.shortcut["system"],
-                     self.menu.dir).replace('/', '\\')
+            args.append(substitute_env_variables(self.shortcut["system"],
```
mbargull: This won't work since the above `cmd` will be used for the `create_shortcut`'s `path` argument, hence the shortcut will be something like `pythonw.exe system_command args...`. Resolution:
1. restore previous behavior, or
2. or make this also use `python cwp env_prefix system_command args` instead of `system_command args`

Without knowing the implications I'd be in favor of the second option since why should the `"system"` case not use the `cwp`-adjusted environment?

mbargull:
BTW (rather off topic and not that important since usually called from GUI/"Start Menu"):
In `cwp.py` the subprocess call is `subprocess.call(args, env=env)`.
Could/should this be replaced by `sys.exit(subprocess.call(args, env=env)` (or `os.execv(args[0], args[1:], env)`?) if it might be used in other contexts (tests or something)?

menuinst/win32.py
msarahan: I don't see this - both are just python.exe here, I think?

msarahan:
@mbargull thanks for the review.  I have tried to make all of your suggested changes.  If you have time, please take one more look.

menuinst/win32.py
mbargull: (omitting paths and file extensions for clarity)
Previously `cmd` was `"python"` (both for main and subprocess). Now `cmd` is `"pythonw"` (from above the `if`) and `subprocess_cmd` is `"python"`.
Thus the shortcuts change from
`python cwp python script`
to
`pythonw cwp python script`
. I don't know if we need to care about this at all... But maybe there is some special Windows voodoo for when to open up the terminal window or how to behave if the linked executable is a console or GUI application -- who knows

mbargull: Ah, I know see that the `"script"` case below uses `u"pythonw.exe"` (even before your changes). Since `"script"` and "`pyscript`" should behave the same concerning terminal windows I'd argue your changes (using `"pythonw"` in any case) should be fine!

menuinst/win32.py
```diff
@@ -251,26 +254,28 @@ def remove(self):
...
+            args = get_python_args_for_subprocess(self.menu.prefix, args, subprocess_cmd)
         elif "pyscript" in self.shortcut:
-            cmd = join(self.menu.prefix, u"python.exe").replace("\\", "/")
+            subprocess_cmd = join(self.menu.prefix, u"python.exe").replace("\\", "/")
             args = self.shortcut["pyscript"].split()
-            args = get_python_args_for_subprocess(self.menu.prefix, args, cmd)
+            args = get_python_args_for_subprocess(self.menu.prefix, args, subprocess_cmd)
         elif "webbrowser" in self.shortcut:
-            cmd = join(unicode_prefix, u'pythonw.exe')
             args = ['-m', 'webbrowser', '-t', self.shortcut['webbrowser']]
```
mbargull: Since the `"system"` case below is now changed to use `cwp`, the same could be done here which in effect makes
```python
args = get_python_args_for_subprocess(self.menu.prefix, args, subprocess_cmd)
```
be done in any case and can thus be done after the if-elif-else-clause to reduce duplicate code :)

mbargull:
Hi @msarahan, what are your thoughts on my last comment on unifying the usage of `cwp.py` for every case, i.e., always properly adjusting the environment variables `PATH`, `CONDA_PREFIX` via `cwp.py`?
Unfortunately, this issue coincided quite precisely with the releases of the last Miniconda/Anaconda installers. But maybe the issue could get resolved in time for the next releases such that a functional `menuinst` will be included with new installations ;P.

goanpeca:
@mbargull you probabluy need to make these changes against 1.4.x not master as master is now tracking 2.x

mingwandroid:
@msarahan, @mbargull, @goanpeca, @nehaljwani, and reason not to get this into the upcoming release?

msarahan:
Switched to target 1.4.x.  This seems to run fine.

@mbargull I don't want to keep cwp.py around, to be honest.  The one true way should be to activate environments, not just set environment variables.  I can try to look into that before the next release, but there might not be time.

goanpeca:
>@mbargull I don't want to keep cwp.py around, to be honest. The one true way should be to activate environments, not just set environment variables. I can try to look into that before the next release, but there might not be time.

Agreed, I am getting rid of cwp on `master` but in the meantime is probably better to include this in 1.4.x

msarahan:
@goanpeca you mean just merge this?  There is middle ground here, I think - cwp.py could call activate?

goanpeca:
>There is middle ground here, I think - cwp.py could call activate?

Hmmm maybe but it feels like an unnecessary complication to do it like that?

mbargull:
@msarahan I made a PR to your branch.
I completely agree `cwp.py` should not be (and not have ever been) part of `menuinst`! This functionality should be incorporated into the development of the environment activation from `conda` itself. The "partial activation" in `cwp.py` diverges from `conda`'s semantics, which is not a good thing.
IMHO it would be nice to have a `conda-run env cmd [args]` command which should be equivalent to something like `(conda activate env && cmd [args])`, i.e. just do the environment activation for `cmd` only and then run it -- BUT omit all the fancy stuff of the old `conda run` command (no need to know about packages etc. at this stage). I guess this could (should?) be developed alongside `conda 4.4`'s new `conda.activate` functionality to maintain `conda`-semantics, unlike `cwp.py`. (Friendly ping @kalefranz ;)). 
@goanpeca 
> Hmmm maybe but it feels like an unnecessary complication to do it like that? if we plan to get rid of it I mean

I agree this is not something that is critical to the next `menuinst 1.4.x` release.

menuinst/win32.py
mingwandroid: I have been reviewing these changes and I don't get what's going on here (not sure if I'm referring to the changes in the PR or some of the stuff we've done previously).

We run both `script` and `system` commands through `pythonw` and `cwp` now which seems wrong to me. We have `script` that contains:
```
        "script": "${PREFIX}/Library/bin/bash.exe",
```
and:
```
        "script": "${PREFIX}/Library/bin/rstudio.exe",
```
.. why would either of these want to be run through `pythonw` and `cwp.py`?

For the `Anaconda Prompt` we have `system` set as:
```
            "system": "%windir%\\System32\\cmd.exe",
            "scriptarguments": ["/K", "${ROOT_PREFIX}\\Scripts\\activate.bat", "${PREFIX}"],
```
.. and again we run that through `pythonw` and `cwp` now!

PR #56, opened 2017-12-19
attempt fix at menuinst logging

kalefranz:


mingwandroid:
I do want it to propagate though, I want to see if with `conda install -v` for example.

PR #61, opened 2018-01-30
more cmd.exe-specific hacking

### LINKED ISSUE #60: Anaconda Prompt broken
from https://github.com/conda/conda/issues/6704#issuecomment-357327468

---

The `Anaconda Prompt` shortcut uses
```cmd
%windir%\System32\cmd.exe "/K" "C:\ProgramData\Anaconda 3\Scripts\activate.bat" "C:\ProgramData\Anaconda 3"
```
for which the last parameter is passed on to `activate.bat` as `"C:\ProgramData\Anaconda 3` (note the missing end quote).
If I change the target to
```cmd
%windir%\System32\cmd.exe /K ""C:\ProgramData\Anaconda 3\Scripts\activate.bat" "C:\ProgramData\Anaconda 3""
```
, everything works fine and `activate.bat`/`conda.bat` get `"C:\ProgramData\Anaconda 3"`.
Conclusion: For the `Anaconda Prompt` this is a `menuinst` (+`console_shortcut`) issue [quoting in `menuinst` is broken](https://github.com/ContinuumIO/menuinst/commit/fcf395dbc7ccfd5130b2c38a243056820034081f)...

----

What we'd actually need, is that in `console_shortcut`
```json
"scriptarguments": ["/K", "${ROOT_PREFIX}\\Scripts\\activate.bat", "${PREFIX}"],
```
should be changed to
```json
"scriptarguments": ["/K", "\"${ROOT_PREFIX}\\Scripts\\activate.bat\" \"${PREFIX}\""],
```
and in `menuinst.win32` *not* add quotes around `/K` via `menuinst.win32.quoted` in https://github.com/ContinuumIO/menuinst/blob/1.4.10/menuinst/win32.py#L292 and *not* let `menuinst.win32.quoted` unconditionally remove quotes at the beginning and end of the arguments in https://github.com/ContinuumIO/menuinst/blob/1.4.10/menuinst/win32.py#L292-L293, cc @mingwandroid. As previously discussed, it would be wise to add a bunch of test cases to `menuinst` to get all those subtleties right; breaking changes might likely have to be postponed for a `menuinst 2.x`, I think.

kalefranz:
fix #60

Believe this should resolve issue #60.  It's very targeted though.  Would like @mingwandroid, @mbargull, @goanpeca to at least be aware of this PR.

### DIFF
--- a/menuinst/win32.py
+++ b/menuinst/win32.py
@@ -123,7 +123,10 @@ def ensure_pad(name, pad="_"):
         '_conda_'
 
     """
-    return name and "%s%s%s" % (pad, name.strip(pad), pad)
+    if not name or name[0] == name[-1] == pad:
+        return name
+    else:
+        return "%s%s%s" % (pad, name, pad)
 
 
 def to_unicode(var, codec=locale.getpreferredencoding()):
--- a/menuinst/win32.py
+++ b/menuinst/win32.py
@@ -123,10 +123,7 @@ def ensure_pad(name, pad="_"):
         '_conda_'
 
     """
-    if not name or name[0] == name[-1] == pad:
-        return name
-    else:
-        return "%s%s%s" % (pad, name, pad)
+    return name and "%s%s%s" % (pad, name.strip(pad), pad)
 
 
 def to_unicode(var, codec=locale.getpreferredencoding()):
--- a/menuinst/win32.py
+++ b/menuinst/win32.py
@@ -115,6 +115,20 @@ def quoted(s):
         return s
 
 
+def ensure_pad(name, pad="_"):
+    """
+
+    Examples:
+        >>> ensure_pad('conda')
+        '_conda_'
+
+    """
+    if not name or name[0] == name[-1] == pad:
+        return name
+    else:
+        return "%s%s%s" % (pad, name, pad)
+
+
 def to_unicode(var, codec=locale.getpreferredencoding()):
     if sys.version_info[0] < 3 and isinstance(var, unicode):
         return var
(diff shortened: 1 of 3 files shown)


menuinst/win32.py
```diff
@@ -115,6 +115,20 @@ def quoted(s):
         return s
 
 
+def ensure_pad(name, pad="_"):
```
kalefranz: I added this function from conda because I don't think `quoted()` above does the right thing.  I definitely don't understand why `u'/' in s` gets quoted.

kalefranz:
CC @msarahan

menuinst/win32.py
```diff
@@ -115,6 +115,20 @@ def quoted(s):
...
+
+    Examples:
+        >>> ensure_pad('conda')
+        '_conda_'
+
+    """
+    if not name or name[0] == name[-1] == pad:
+        return name
+    else:
+        return "%s%s%s" % (pad, name, pad)
```
mingwandroid: How will this behave when it sees a single pad at the start or end? It'll double that one up. Is that what you want?

kalefranz: > Is that what you want?

I think that's what I wanted, yes.  Let me look again at this code now that I have the activate.bat hack 🤮 done.

kalefranz: I don't think it matters in most cases.  Probably only matters in weird edge cases.  The old version of `ensure_pad()` (from conda, see https://github.com/conda/conda/pull/6817) did what you think it should.  When I was writing this, I got scared `ensure_pad()` was doing something it shouldn't.  I guess what you *want* for behavior can be ambiguous here.  But I'll go back to the original.

kalefranz:
CC @jjhelmus

menuinst/win32.py
```diff
@@ -115,6 +115,17 @@ def quoted(s):
...
 
+def ensure_pad(name, pad="_"):
+    """
+
+    Examples:
+        >>> ensure_pad('conda')
+        '_conda_'
+
+    """
+    return name and "%s%s%s" % (pad, name.strip(pad), pad)
```
jjhelmus: This will replace multiple instances of  pad  with a single instance, which will be an issue if the arguments are already quoted to multiple levels:

``` Python
>>> "_a_".strip("_")
'a'
>>> "__a__".strip("_")
'a'
```

Trying to figure out a nice solution to this.

kalefranz: In the first commit on this PR I had

```python
if not name or name[0] == name[-1] == pad:
    return name
else:
    return "%s%s%s" % (pad, name, pad)
```

jjhelmus: How about the following:

``` Python
def ensure_pad(s, pad="_"):
    if s[0] != pad:
        s = pad + s
    if s[-1] != pad:
        s += pad
    return s
```

The only question is if created an unbalanced number of quotes will create an issue, which I think it will

jjhelmus: I would go with the original:

```
if not name or name[0] == name[-1] == pad:
    return name
else:
    return "%s%s%s" % (pad, name, pad)
```

That will never add unbalanced quotes and doesn't strip existing cases of multiple quotes

kalefranz: Ok original added back.  And also re-opened https://github.com/conda/conda/pull/6817 because I want to keep these two functions consistent.

mingwandroid:
Merging, thanks @kalefranz and @jjhelmus. @csoja, we need to bump the version number and do a release still.

PR #62, opened 2018-02-06
fix #54 incorrect locale handling

### LINKED ISSUE #54: windows.knownfolders.get_path uses incorrect locale handling
At [knownfolders.py#L213](https://github.com/ContinuumIO/menuinst/blob/master/menuinst/windows/knownfolders.py#L213) we see this code:

```python
    # convert to unicode
    codec = sys.getdefaultencoding()
    if not codec:
        codec="utf-8"
    if hasattr(path, "decode"):
        path = path.decode(codec)
```

This entire block is incorrect and unnecessary, and actually broken on Python 2:
* `sys.getdefaultencoding()` is for transparently converting between `str` and `unicode` (**`ascii` on Python 2!**) - you actually want `sys.getfilesystemencoding()`
* the `ctypes` code above uses `wchar_t`, so `path` is already `str` on Python 3.x and `unicode` on Python 2.x
* on Python 3.x this is a no-op (`hasattr(str(), "decode")` is always `False`)
* on Python 2.x this throws if `path` has "invalid" characters, but by definition it cannot have invalid characters because it has come from the OS. Without changing the codec, "invalid" means "non-ASCII", which is _very_ restrictive

The end result is that if you invoke this code path - for example, by installing Anaconda2 with non-ASCII characters in your username - you will get a `UnicodeEncodeError` while doing an unnecessary decode.

Removing the entire quoted block above is the simplest and most correct fix.

kalefranz:
fix #54

### DIFF
--- a/.cache/v/cache/lastfailed
+++ b/.cache/v/cache/lastfailed
@@ -1,0 +1,3 @@
+{
+  "tests/test_win32.py": true
+}
--- a/menuinst/knownfolders.py
+++ b/menuinst/knownfolders.py
@@ -198,12 +198,6 @@ def get_path(folderid, user_handle=UserHandle.common):
 
     path = pPath.value
     _CoTaskMemFree(pPath)
-    # convert to unicode
-    codec = sys.getdefaultencoding()
-    if not codec:
-        codec="utf-8"
-    if hasattr(path, "decode"):
-        path = path.decode(codec)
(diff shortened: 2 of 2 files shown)


kalefranz:
@mingwandroid Are you ok with merging this for the release also?  @nehaljwani confirmed it looks correct to him.

PR #63, opened 2018-02-06
fix logging traceback from #54

kalefranz:
fixes comment from https://github.com/ContinuumIO/menuinst/issues/54#issuecomment-349554041


kalefranz:
I don't see any other similar issues in win32.py

PR #67, opened 2018-05-08
put recursion guard in install func

### LINKED ISSUE #42: conda uncontrollably spawning python processes when creating/removing envs
_From @cwreid18 on March 2, 2017 16:48_

On Windows 7, with Anaconda installed for all users in C:\Anaconda3. No problem creating and removing envs with admin rights. However, when running as user without admin rights, I get the following problem when creating environments: packages are downloaded ok, but then it hangs (usually at the linking or unlinking packages step) and I can see pythonw.exe processes being launched in task manager at a rate of several per second. Only way to stop this is to shut down the machine. The problem is intermittent and seems to occur more frequently when creating envs with all the anaconda packages (e.g. conda create -n test anaconda=4.1.1) rather than environments that only contain a few packages. Happens occasionally when trying to remove envs. I have tried reinstalling with multiple versions of anaconda (inc. the latest, 4.3.0) but the problem keeps occurring.

_Copied from original issue: conda/conda#4782_

msarahan:
fixes #42 

CC @wulmer @asishm @j-hartshorn @Ritcho @plu2k @jengelman

I have been unable to reproduce the issue locally.  However, I think I understand it thanks to your descriptions.  I believe that this was an infinite loop within the install process.  The code here should prevent that.

If you can test this on your end, I'd be very thankful.

### DIFF
--- a/menuinst/__init__.py
+++ b/menuinst/__init__.py
@@ -48,17 +48,22 @@ def _install(path, remove=False, prefix=sys.prefix, mode=None):
             ShortCut(m, sc).create()
 
 
-def install(path, remove=False, prefix=sys.prefix):
+def install(path, remove=False, prefix=sys.prefix, recursing=False):
     """
     install Menu and shortcuts
     """
     if sys.platform == 'win32' and not exists(join(sys.prefix, '.nonadmin')) and not isUserAdmin():
         from pywintypes import error
         try:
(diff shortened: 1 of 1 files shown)


msarahan:
Proceeding.  If anyone has troubles like this again, let me know.

PR #73, opened 2019-03-06
Command line options starting with minus, or a space shall not be quo…

katietz:
Don't add quotes for command line options starting with a space, or with a minux.

### DIFF
--- a/menuinst/windows/win32.py
+++ b/menuinst/windows/win32.py
@@ -113,6 +113,9 @@ def quoted(s):
     """
     # strip any existing quotes
     s = s.strip(u'"')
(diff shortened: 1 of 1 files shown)


menuinst/windows/win32.py
```diff
@@ -113,6 +113,9 @@ def quoted(s):
     """
     # strip any existing quotes
     s = s.strip(u'"')
+    # don't add quotes for minus or leading space
+    if u'-' in s[:1] or u' ' in s[:1]:
```
msarahan: more compact, pythonic way to write this:

```suggestion
    if s[0] in (u'-', u' '):
```

PR #85, opened 2021-08-02
Include patches used in conda-standalone

ericpre:
- [x] includes patches used in https://github.com/conda-forge/conda-standalone-feedstock and https://github.com/AnacondaRecipes/conda-standalone-feedstock 
- [x] resolve conflict with menuinst 1.4.17

These two commits are used as patch in https://github.com/conda-forge/conda-standalone-feedstock/pull/18.

### DIFF
--- a/menuinst/__init__.py
+++ b/menuinst/__init__.py
@@ -24,8 +24,8 @@ elif sys.platform == 'win32':
     from .win_elevate import isUserAdmin, runAsAdmin
 
 
-def _install(path, remove=False, prefix=sys.prefix, mode=None):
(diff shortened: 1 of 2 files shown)


jezdez (changes requested):
This seems certainly like a good feature, but it'd be splendid to have test coverage for the new parameter. Do you think the current tests are capable of handling this?

ericpre:
I added tests, which create conda environment to check the existence of the shortcut and its name.

ericpre:
@jezdez, could you please have a look at this PR? Thanks!

PR #88, opened 2021-10-08
fix support for osx

jaimergp:
This PR adds the minimum amount of changes needed to make OSX applications work again. It's not the most elegant code, but let's start small.

Summary of changes:

* Recover `utils.get_executable` from the original `appinst`
* Fix class signatures
* Conform to the `menuinst.install(...)` API expected by `conda` (same as windows)
* Comply with the modern `plistlib` API
* Slugify script name (last component of`/Applications/${NAME}.app/Contents/MacOS/${NAME}`) 
* Bring some placeholders used on Windows too

### DIFF
--- a/menuinst/darwin.py
+++ b/menuinst/darwin.py
@@ -5,16 +5,28 @@
 import os
 import sys
 import shutil
-from os.path import basename, join
-from plistlib import Plist, writePlist
+from os.path import basename, join, expanduser
+import plistlib
+import logging
 
-from utils import rm_rf
+from .utils import rm_rf, slugify
+
+logger = logging.getLogger("menuinst_osx")
+logger.setLevel(logging.DEBUG)
+stream_handler = logging.StreamHandler()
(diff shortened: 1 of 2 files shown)


hoechenberger:
I'd just like to say that it would be great to have these changes added to `menuinst`. If there's any way in which I could help make that happen, please do let me know.

isuruf:
This looks good to me. Can we enable tests for this so that we don't accidentally break this in the future?

jaimergp:
Thanks @isuruf! I don't know if we want to pursue this series of updates though (this PR, and the one for Linux).

There's a full rewrite on #91 that implements the same functionality with a new cross-platform schema. It could be considered a new major version, though.

Enabling non-Windows also requires changes in `conda` itself. 

I think we should close this PR, and #89, and work our way towards cross-platform compatibility through #91.

hoechenberger:
> I think we should close this PR, and #89, and work our way towards cross-platform compatibility through #91.

I agree!

PR #103, opened 2022-08-10
Copy ShellExecuteEx from conda to replace pywin32

### LINKED ISSUE #41: pywin32 dependency / size of menuinst in Miniconda etc.
Since #27 `win_elevate.py` is included, which is great since it resolved interactive elevation. This added `pywin32` as a dependency though.
Since `pywin32` is a rather "heavy" package, now `menuinst` (with its dependency) will account for a considerable size of minimal installers like Miniconda.
Sadly, I do not know of an alternative to `pywin32` which provides similar functionality for Windows' interactive UAC.

isuruf:
Remove the pywin32 dependency which does not build on pypy

### Checklist - did you ...

cc @mbargull, @mattip, @h-vetinari

Fixes #41

### DIFF
--- a/menuinst/win_elevate.py
+++ b/menuinst/win_elevate.py
@@ -10,6 +10,7 @@
 
 from __future__ import print_function
 import sys, os, traceback
+from enum import IntEnum
 
 if sys.version_info < (3,):
     text_type = basestring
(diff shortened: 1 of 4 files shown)


isuruf:
@conda/conda-core, this is ready for a review. There's no news folder as the PR template suggests.

mattip:
cool. This is similar to jupyter/jupyter_core#230

menuinst/win_elevate.py
```diff
@@ -35,42 +36,119 @@ def isUserAdmin():
         raise RuntimeError("Unsupported operating system for this module: %s" % (os.name,))
 
 
+# Taken from conda/common/_os/windows.py
+if os.name == 'nt':
+    from conda.common.compat import ensure_binary
```
jaimergp, line 41: I think we need to copy this simple function over too, so we don't depend on `conda`:

```python
def ensure_binary(value):
    try:
        return value.encode('utf-8')
    except AttributeError:  # pragma: no cover
        # AttributeError: '<>' object has no attribute 'encode'
        # In this case assume already binary type and do nothing
        return value
```

isuruf: Good catch. Fixed now.

jaimergp:
> There's no news folder as the PR template suggests.

I guess we can update the `CHANGES.txt` file for now.

jaimergp (approved):

hoechenberger (approved):

jaimergp (approved):

h-vetinari:
Is there a rough timeline for a release that includes this change, or do we need to backport this in conda-forge? would be good if we could move on with removing pypy3.7 one way or another.

h-vetinari:
PS: thanks a lot for finding a work-around for this!

hoechenberger:
@h-vetinari I don't think we've agreed on a release schedule, but with all the numerous improvements that @jaimergp and others have recently merged, I'd say it'll soon be time for a new major release!!

jaimergp:
I guess we can cut a release for menuinst and reissue a conda-standalone build too.

isuruf:
Sounds good. Who can make a release? I guess this can be just a patch/minor version bump.

jaimergp:
I'll do it. I think we just need to tag.

PR #133, opened 2023-06-19
Set `AppUserModelId` on Windows shortcuts

### LINKED ISSUE #127: Set app ID on Windows
### Checklist

### What is the idea?

Comes from https://github.com/spyder-ide/spyder/issues/20791#issuecomment-1505877877

---

Set some sort of application ID on Windows shortcuts so icons are properly grouped when needed.

### Why is this needed?

_No response_

### What should happen?

We need to add:

- A new field on the schema, with validation according to [this info](https://learn.microsoft.com/en-us/windows/win32/shell/appids#how-to-form-an-application-defined-appusermodelid)
- Some code on the C extension for windows shortcuts. Essentially, fetch the property store for the shortcut and set the adequate key-value pair. (see Resources below)

### Additional Context

Useful resources:

- https://github.com/search?q=PKEY_AppUserModel_ID+language%3Ac&type=code
- https://nsis.sourceforge.io/WinShell_plug-in
- https://learn.microsoft.com/en-us/windows/win32/shell/appids
- https://learn.microsoft.com/en-us/windows/win32/properties/props-system-appusermodel-id

jaimergp:
### Description

Will close #127 

### Checklist - did you ...

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -81,7 +81,7 @@ jobs:
         name: Run test suite
         run: |
           python -I -m pytest tests/ --cov-append --cov-report=xml --cov=menuinst -vvv
-      
+
       - name: Setup tmate session
         if: ${{ failure() }}
         uses: mxschmitt/action-tmate@v3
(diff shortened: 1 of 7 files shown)


jaimergp:
@mrclary - this looks like it's working. Before I create some dev packages for us to try on installers, can you take a look a this PR and let me know if there's anything off? Thanks!

mrclary:
@jaimergp, I cannot comment on the C++ code, but everything else looks good to me. I probably won't get a chance to test the branch until tomorrow, but I don't think you need wait on me.

Issue #137, opened 2023-06-21
MENU_ITEM_LOCATION is incorrect for all-user installation on macOS


mrclary:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

@jaimergp, when selecting "Install on a specific disk..." from a package installer created by `constructor` (to install for all users) on macOS, the environments are created in the correct place but the application bundle is created in `~/Applications` instead of `/Applications`.

mrclary:
@jaimergp, perhaps this is a bug in `constructor`? Maybe it's not supplying `_mode='system'` to `menuinst.api.install`?

jaimergp:
Hey! Thanks for the report. Technically, `menuinst` should "detect" whether the installation is for a single-user or not via the presence of `.nonadmin` file in the conda installation root ([see this block](https://github.com/conda/menuinst/blob/d374862ae6cd038638653a9c706b1b7001ef757d/menuinst/utils.py#L326-L333)).

This used to be Windows-specific and I don't think I have added this logic in `constructor`, so the bug is indeed in `constructor` 😬  We need figure out a way to detect whether the installation is single-user or all users and add the `.nonadmin` file accordingly. 

I'll mention the issue in the relevant PR there, thanks!

jaimergp:
@mrclary - this commit should be enough to create the `.nonadmin` files: https://github.com/conda/constructor/pull/474/commits/3f0b97bfc5b43441f2271b4c67df3c952d078e16.


However, thinking about it, they were _not_ being created before, which means that something is also off in `menuinst` (which should have understood that the installations were admin-like).

mrclary:
https://github.com/conda/menuinst/blob/237e259a752e49d38bb256135ade6ce9ada36e01/menuinst/_legacy/win32.py#L140-L142

I found this _only_ in the legacy code for Windows. Nothing like this in `platforms/base.py:Menu` or the platform modules. I have not tested admin install on Linux, and I don't recall checking the shortcut location for the admin install for Windows. So it is possible that shortcut locations for all platforms are created in a user-specific location...

jaimergp:
It's here, in the `elevate_as_needed` decorator (which I want to rewrite anyway):

https://github.com/conda/menuinst/blob/237e259a752e49d38bb256135ade6ce9ada36e01/menuinst/utils.py#L326

mrclary:
Aahh, then perhaps this preceding clause is the problem?
[...]

PR #138, opened 2023-06-29
Enable elevation on non-Windows platforms

jaimergp:
### Description

Closes https://github.com/conda/menuinst/issues/137

### Checklist - did you ...

### DIFF
--- a/menuinst/utils.py
+++ b/menuinst/utils.py
@@ -375,12 +375,18 @@ def elevate_as_needed(func: Callable) -> Callable:
     return wrapper_elevate
 
 
-def _test(base_prefix: Optional[os.PathLike] = None, _mode: str = "user"):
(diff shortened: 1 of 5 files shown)


menuinst/utils.py
```diff
@@ -352,6 +350,7 @@ def wrapper_elevate(
                         f"{func.__name__}("
                         f"*{args!r},"
                         f"base_prefix={base_prefix!r},"
+                        f"_mode='system',"
```
jaimergp, line 353: @mrclary - this was the bug 😬

mrclary: ... when `.noadmin` is absent _and_ `user_is_admin()` is False. So when a non-admin user runs `menuinst` from the commandline? When `constructor` runs for all users, `user_is_admin()` should return True, correct?

PR #147, opened 2023-08-15
Fix segfaults on Windows while building conda packages

jaimergp:
### Description

Windows canaries failed before hitting the test phase with a segfault, possibly caused by the presence of OS-foreign binaries in the package. Deleting them before running the install (since they are not used on Windows anyway) does the trick.

### Checklist - did you ...

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -13,6 +13,7 @@ build:
     # Apparently these files make the post-build linkage analysis crash
     # and we should not need them on Windows
     - del /q "{{ SRC_DIR }}\\menuinst\\data\\osx_launcher_*"  # [win]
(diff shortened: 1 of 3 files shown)


recipe/meta.yaml
```diff
@@ -13,6 +13,7 @@ build:
     # Apparently these files make the post-build linkage analysis crash
     # and we should not need them on Windows
     - del /q "{{ SRC_DIR }}\\menuinst\\data\\osx_launcher_*"  # [win]
+    - del /q "{{ SRC_DIR }}\\menuinst\\data\\appkit_launcher_*"  # [win]
```
jaimergp, line 16: conda-build post-build linkage analysis doesn't like unix binaries (at least these) when running on Windows 🤷

jezdez: Huh TIL!

jezdez (approved):

PR #156, opened 2023-09-06
Rework automatic elevation logic

### LINKED ISSUE #150: `.nonadmin` files in pre-menuinst v2 conda installations
> This is an issue for both `constructor` and `menuinst` I guess, but I'll keep it here because I think it has to do more with `menuinst` (even if the fix is in `constructor` proper).

---

The context:

* On Windows, `menuinst` v1 has some [super user permissions elevation logic](https://github.com/conda/menuinst/blob/1.4.x/menuinst/win_elevate.py) to enable the creation of shortcut items in the Start Menu for "All Users" installations.
* The way `menuinst` anticipates whether it needs superuser permissions or not is by checking the presence of a `.nonadmin` file in the root conda installation (aka `%PREFIX%\.nonadmin`). The presence of the file dictates that no extra permisions are needed. The absence means superuser was needed at installation time.
  * This file is [placed there by the constructor-based installer](https://github.com/conda/constructor/blob/0cc258e2012f43567e46504ad377cc8214ef5405/constructor/nsis/main.nsi.tmpl#L1100-L1104) when the user installs Miniconda, Miniforge, etc. 
  * It is primarily there to notify the uninstaller whether superuser is needed or not, but menuinst also uses it because well, it's there.
* Again, this is a Windows-only thing! `.nonadmin` doesn't exist in Linux or macOS, even if no special permissions were needed to install `conda`.
 
The problem:

- `menuinst` v2 extends shortcut creation to all platforms, Linux and macOS included.
- It still checks for permissions needed or not via the presence of this file.
- However, existing Unix installations (all of those provided by a constructor-based installer built with constructor 3.4.5 or earlier), will NOT have the `.nonadmin` file even if no special permissions were needed.
- This will make `menuinst` believe that, since `.nonadmin` is absent, it indeed NEEDS superuser permissions, and will request `sudo` powers, when it doesn't technically need them.

Possible solutions:

- Add a `post-link.sh` script to the `menuinst` v2 package to add the `.nonadmin` file to `$CONDA_ROOT` in Unix systems. The logic would be something like "if it's not there, try to add it; if we could do it, it was a nonadmin install; if we failed, that's ok, it was an admin install".
- Change how we approach the "superuser permissions needed" on Unix systems. Maybe we can ignore the `.nonadmin` file and instead check whether the `CONDA_ROOT` path is in a system-wide location (`/opt`, `/usr/local`...), outside of the current user home (`~/`), while handling if we are sudo already or not. This might have a few rough edges though; e.g. what if the user installed to a non-standard location like `/my-stuff/conda`. Is that user-only? All users? Maybe it's world writable but, does that mean that the shortcuts I am processing should be available to all users?

---

I feel we need to rethink how this is approached outside Windows, specially because HPC users will probably run Linux and in those setups you can find system-wide conda installations with per-user environments that should NOT ask for superuser permissions to process their own shortcuts. 

Maybe it's all way simpler: if the installation is running with superuser permissions already, then install to an "All users" location. If it's not, we'll install to current user directories. But we should not depend on the state of the `$CONDA_ROOT` folder.

Thoughts? Tagging a few folks that have shown interest in the project or the superuser elevation issue before: @aganders3 @mrclary @marcoesters. Thanks!

jaimergp:
### Description

Closes #150 

### Checklist - did you ...

### DIFF
--- a/menuinst/utils.py
+++ b/menuinst/utils.py
@@ -3,10 +3,9 @@ import re
 import shlex
 import subprocess
 import sys
-import traceback
 import xml.etree.ElementTree as XMLTree
 from contextlib import suppress
-from functools import wraps
+from functools import wraps, lru_cache
 from logging import getLogger
 from pathlib import Path
 from typing import Callable, Iterable, Literal, Mapping, Optional, Sequence, Union
@@ -246,6 +245,52 @@ def deep_update(mapping: Mapping, *updating_mappings: Iterable[Mapping]) -> Mapp
     return updated_mapping
 
 
+def needs_admin(prefix: os.PathLike, base_prefix: os.PathLike) -> bool:
+    """
+    Checks if the current installation needs admin permissions.
+    """
+    if user_is_admin():
+        return False
+    
+    if Path(prefix, ".nonadmin").exists():
+        # This file is planted by the constructor installer
+        # and signals we don't need admin permissions
+        return False
+    
+    try:
+        Path(prefix, ".nonadmin").touch()
+        return False
+    except Exception as exc:
+        logger.debug("Attempt to write %s/.nonadmin failed.", prefix, exc_info=exc)
+
+    if base_prefix == prefix:
+        # We are already in the base env, no need to check further
+        return True
+
+    # I can't think of cases where users can't write to the target prefix but can to base
+    # so maybe we can skip everything underneath?
+
+    if Path(base_prefix, ".nonadmin").exists():
+        return False
+
+    if os.name == "nt":
+        # Absence of $base_prefix/.nonadmin in Windows means we need admin permissions
+        return True
+    
+    if os.name == "posix":
+        # Absence of $base_prefix/.nonadmin in Linux, macOS and other posix systems
+        # has no meaning for historic reasons, so let's try to see if we can
+        # write to the installation root
+        try:
+            Path(base_prefix, ".nonadmin").touch()
+        except Exception as exc:
+            logger.debug("Attempt to write %s/.nonadmin failed.", prefix, exc_info=exc)
+            return True
+        else:
+            return False
+
+
+@lru_cache(maxsize=1)
 def user_is_admin() -> bool:
     if os.name == "nt":
         from .platforms.win_utils.win_elevate import isUserAdmin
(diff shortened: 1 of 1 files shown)


jaimergp:
Tagging @aganders3 @mrclary @marcoesters in case they are interested in a quick review, as discussed in #150 🙏

menuinst/utils.py
```diff
@@ -246,6 +245,52 @@ def deep_update(mapping: Mapping, *updating_mappings: Iterable[Mapping]) -> Mapp
...
+        return False
+
+    if Path(target_prefix, ".nonadmin").exists():
+        # This file is planted by the constructor installer
+        # and signals we don't need admin permissions
+        return False
+
+    try:
+        Path(target_prefix, ".nonadmin").touch()
+        return False
```
marcoesters, line 262: The assumption here is that if I can write into the directory, this is a non-admin installation to begin with, that file just hasn't been written before for some reason?

The `user_is_admin` check should protect from any kind of problematic permanent changes, but I'm wondering if that file should be cleaned up afterwards.

jaimergp: Yep, we should only expect a `.nonadmin` file in the base environment of Windows installations. It won't exist anywhere else. What we are doing here is to try to create that file in the target environment (because that's where the menuinst JSON came from) for the first time. If we succeed, then we must be running in user mode, and we leave it there so we don't have to check in future runs. We do assume that permissions won't change in the future, and that other users will not try to write to it if they don't have permissions (I'm thinking HPC setups with an admin-protected base, but user-writable envs in their home folders).

marcoesters: > We do assume [...] that other users will not try to write to it if they don't have permissions 

That's why I'm wondering if we should remove touched files to be safe

jaimergp: Maybe only if they are old enough? Let's say at least a day old? Otherwise we will be creating and recreating files everytime `menuinst` is run, which can be several times per `conda` command.

mrclary:
> Tagging @aganders3 @mrclary @marcoesters in case they are interested in a quick review, as discussed in #150 🙏

LGTM.

menuinst/utils.py
marcoesters: After thinking about this more, I think the risk is minimal. If you can write to the directory during the installation, you can do much worse than create shortcuts. And the presence of a `.noadmin` file really only tells `menuinst` to attempt to do a non-admin installation

marcoesters (approved):

aganders3:
> Tagging @aganders3 @mrclary @marcoesters in case they are interested in a quick review, as discussed in https://github.com/conda/menuinst/issues/150 🙏

I'm not quite as familiar with this code, but these changes make sense to me!

jaimergp:
Thanks everyone!

PR #160, opened 2023-09-15
add note about swift and macos 10.14.4+

jaimergp:
### Checklist - did you ...

Comes from discussion at https://github.com/conda-forge/menuinst-feedstock/pull/23#discussion_r1326332333

### DIFF
--- a/docs/source/defining-shortcuts.md
+++ b/docs/source/defining-shortcuts.md
@@ -66,6 +66,8 @@ Unix systems have the notion of MIME types, while Windows relies more on file na
   files) placeholders to your command so the paths are passed adequately. If you are defining a new
   MIME type, you must fill the `glob_patterns` field by mapping the new MIME type to the file
   extensions you want to associate with it.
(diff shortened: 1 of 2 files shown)


docs/source/defining-shortcuts.md
```diff
@@ -144,6 +153,14 @@ shortcut.
   communication mechanism). Note that setting `CFBundleTypeRole` will make the wrapper blip in the
   dock when the URL is opened. If you don't want that, do not set it.
 
+:::{note}
```
travishathaway, line 156: Why is this note included twice so close to the other one? Mistake?

jaimergp: Ah, it's the same note for two related features: associating file extensions and URL protocols. In macOS, these are handled by the same system, but it's not the same in other platforms. So just to make sure folks do see it if they are only handling one of the cases.

travishathaway: Okay, sounds good to me.

travishathaway (approved):

PR #163, opened 2023-10-23
Add Python 3.12 to CI matrix

xhochy:
### Checklist - did you ...

- [ ] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [ ] Add / update necessary tests?
- [ ] Add / update outdated documentation?

PR #164, opened 2023-10-25
Add support for Python 3.12

jaimergp:
### Description

Supersedes https://github.com/conda/menuinst/pull/163. Apparently the C extension on Windows does not compile for 3.12 ([see issues in feedstock](https://github.com/conda-forge/menuinst-feedstock/pull/31)).

### Checklist - did you ...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,9 @@
 * @dbast made their first contribution in https://github.com/conda/menuinst/pull/116
 * @aganders3 made their first contribution in https://github.com/conda/menuinst/pull/119
 
+## 1.4.20 (2023-10-30)
+
+  * Add support for Python 3.12 (#165)
 
 ## 1.4.19 (2022-08-17)
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -65,7 +65,6 @@ jobs:
           conda install -yq --solver libmamba \
             --file tests/requirements.txt \
             python=${{ matrix.python-version }}
-          conda install -yq --solver libmamba conda --only-deps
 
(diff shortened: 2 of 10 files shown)


jaimergp:
Not seeing the same errors as in the feedstock, as it compiles, but the extension does fail to run later in the tests:

```
            # winshortcut is a windows-only C extension! create_shortcut has this API
            # Notice args must be passed as positional, no keywords allowed!
            # winshortcut.create_shortcut(path, description, filename, arguments="",
            #                             workdir=None, iconpath=None, iconindex=0, app_id="")
>           create_shortcut(
                target_path,
                self._shortcut_filename(ext=""),
                str(path),
                " ".join(arguments),
                working_dir,
                icon,
                0,
                self._app_user_model_id(),
            )
E           SystemError: argument 1 (impossible<bad format char>)
```

jaimergp:
Ok, that seems to work! Errors are due to activation not working because we are installing conda in a non-conventional way to bypass the cyclical dependency between conda and menuinst.

@xhochy, if you have some time, can you take a look at the code? I'm mostly pushing buttons til it works here 😂

jaimergp:
Can I get a review here @conda/conda-core? It's essentially the same as https://github.com/conda/menuinst/pull/165, and it's blocking https://github.com/conda/conda/pull/13073

CHANGELOG.md
kenodegard, line 1: Do we need to manually modify the CHANGELOG now that we use rever?

kenodegard: ah i see, this is effectively just merging `1.4.x` into `main`

travishathaway (approved):

PR #165, opened 2023-10-26
[1.4.x] Improve support for Python 3.12

jaimergp:
### Description

1.4.x backport of #164 plus other fixes required for Python 3.12

### Checklist - did you ...


jaimergp:
@xhochy, we can try this backport directly on the feedstock as a [patch](https://patch-diff.githubusercontent.com/raw/conda/menuinst/pull/165.patch). If that works, we can cut a patch release for the 1.4.x branch.

menuinst/winshortcut.cpp
```diff
@@ -30,18 +32,24 @@
 
 static PyObject *CreateShortcut(PyObject *self, PyObject *args)
 {
-    Py_UNICODE *path; /* path and filename */
-    Py_UNICODE *description;
-    Py_UNICODE *filename;
+    PyObject *py_path; /* path and filename */
```
xhochy, line 35: Don't we need to free (as in `Py_DECREF `) these objects at the end?

jaimergp: I tried in #164 but it crashes. According to @ngoldbaum, these are borrowed references so they don't need decref'ing.

xhochy (approved):
Looking good!

menuinst/winshortcut.cpp
ngoldbaum: in the docs for PyArg_ParseTuple, under the `O` code:

> The C program thus receives the actual object that was passed. A new strong reference to the object is not created (i.e. its reference count is not increased).

`U` is the same as `O` except is does a PyUnicode_Check as well.

jezdez (approved):

PR #180, opened 2024-01-25
Support different names for base and non-base environment

marcoesters:
### Description

`menuinst v1` added the environment name to shortcuts when the packages are installed outside the base environment. ~This behavior appears to be intended for Windows: https://github.com/conda/menuinst/blob/1df31c6ad71110fb9d037b9d6aca5b35da08d0c3/menuinst/platforms/win.py#L201~

~However, `env_name` is never set. This PR sets `env_name` outside the base environment.~

Add a feature to re-enable this behavior by expanding the schema. Contrary to v1, this is an option now and not standard behavior to allow for more flexibility.

### Checklist - did you ...


jaimergp:
I was thinking about the setup we use in napari. The installers basically ship:

- A `base` environment with conda and friends
- A `napari-X.Y.Z` environment, where X.Y.Z is the version of napari.

If we merge this PR, napari shortcuts will _always_ have that env name appended, which might be redundant (we will see something like `napari (napari-0.4.19)`). I wonder if this is something we should control at the menuinst level or the constructor level. Maybe it should be a special placeholder in the shortcut name instead? So we make it opt-in?

Something like:

```json
...
   "name": "Spyder 4.0 {{ ENV_NAME_IF_NOT_BASE }}",
...
```

Too hacky?

ccordoba12:
> Too hacky?

We are talking with @mrclary to do exactly the same for Spyder! So, if you decide to implement that directly here, we wouldn't need to in our `menu.json`

marcoesters:
I don't think it's too hacky. This is exactly the behavior that many seem to want.

We could alternatively also define the `ENV_NAME` that way - that it only renders if it's not the `base` environment. However, `ENV_NAME_IF_NOT_BASE` would maintain backwards compatibility with other patch versions of version 2.

mrclary:
> We are talking with @mrclary to do exactly the same for Spyder! So, if you decide to implement that directly here, we wouldn't need to in our `menu.json`

I'm sorry, @ccordoba12, I need to clarify that this is not exactly what we are trying to do with Spyder. What we are doing for Spyder is setting the name of the shortcut as `Spyder <major version>` for our *installer created* (non-base) environment, and `Spyder <major version> (<env name>)` for any other conda environment.

I do not support this PR as it currently is because I don't think `menuinst` should _automatically_ set any part of the shortcut name, removing flexibility from the developer. This PR would absolutely prevent us from doing what we want to do with Spyder. I *would* support additional `menuinst` placeholders or additional `menuinst._schema.MenuItem`s, e.g. in addition to `name`, have `name_if_base` and/or `name_if_not_base`, or similar, with defined precedence.

These would not really do anything for Spyder since `menuinst` has no way of knowing whether the environment was created by `constructor`, but they would not *prevent* us from doing it. So we would still use `constructor`'s `pre-install` script in combination with `spyder`'s `post-link` script to distinguish the `constructor` made environment and modify the `spyder-menu.json` before `menuinst` is invoked.

marcoesters:
I think a separate menu item is a better way to go about this. A separate placeholder would still require implicit `menuinst` logic. For example:

```
name: App {{ NAME_IF_NOT_BASE }}
```

would either have to implicitly insert parentheses or would just insert the name without them. Adding an optional `name_if_not_base` provides more flexibility.

jaimergp:
The more I think about this the less I see a way out, hah. `menuinst` is supposed to be not too conda specific. Other package managers could implement hooks to use it with different JSON locations. For example, `micromamba` or `pixi` could implement the schema and over there there's no concept of `base` environment.

I'm not sure I like the idea of adding more items to an implementation agnostic schema, when those items are kind of implementation specific... I need to think more about this 😬

marcoesters:
`menuinst` does already have the concept of `prefix` and `base_prefix` though. RIght now, it's only `conda`, but `utils._default_prefix` could be expanded to accommodate other frameworks.

If they do not have a concept like this, they don't have to use these, i.e., `prefix` and `base_prefix` will be the same thing.

ccordoba12:
> I'm sorry, @ccordoba12, I need to clarify that this is not exactly what we are trying to do with Spyder

Ok, sorry for the misunderstanding.

> I do not support this PR as it currently is because I don't think menuinst should automatically set any part of the shortcut name

You're probably right: any project that depends on `menuinst` should be able to decide how to manage to name their shortcuts.

jaimergp:
What if we extend `name` field into a multi-type field. Right now, it only takes a `str` (with placeholders), but we could also allow condition-to-str mappings:

```json
...
"name": {
  "prefix_is_base": "Something",
  "prefix_is_not_base": "Something (else)"
},
...
```

I think I like that better, and it's a bit more future proof because if there are more use cases in the future, we won't have to add yet another key, only more conditions to the `name` mapping.

mrclary:
> What if we extend `name` field into a multi-type field. Right now, it only takes a `str` (with placeholders), but we could also allow condition-to-str mappings:
> 
> ```json
> ...
> "name": {
>   "prefix_is_base": "Something",
>   "prefix_is_not_base": "Something (else)"
> },
> ...
> ```
> 
> I think I like that better, and it's a bit more future proof because if there are more use cases in the future, we won't have to add yet another key, only more conditions to the `name` mapping.

@jaimergp, I like this idea.
"prefix_is_base" and "prefix_is_not_base" may be a bit confusing, however. "prefix" and "base" may be mis-interpreted as name prefix and base name, rather than referring to the environment prefix and base environment. Perhaps keys such as "default" and "base_env" might be more clear. Users could specify just "default" and have it apply everywhere, but if "base_env" is specified, then that value is applied in the case of a base environment.

marcoesters:
> > What if we extend `name` field into a multi-type field. Right now, it only takes a `str` (with placeholders), but we could also allow condition-to-str mappings:
> > ```json
> > ...
> > "name": {
> >   "prefix_is_base": "Something",
> >   "prefix_is_not_base": "Something (else)"
> > },
> > ...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think I like that better, and it's a bit more future proof because if there are more use cases in the future, we won't have to add yet another key, only more conditions to the `name` mapping.
> 
> @jaimergp, I like this idea. "prefix_is_base" and "prefix_is_not_base" may be a bit confusing, however. "prefix" and "base" may be mis-interpreted as name prefix and base name, rather than referring to the environment prefix and base environment. Perhaps keys such as "default" and "base_env" might be more clear. Users could specify just "default" and have it apply everywhere, but if "base_env" is specified, then that value is applied in the case of a base environment.

Sorry for leaving this dormant - just getting back to this after a long leave of absence.

I do like the mapping idea, but we need to make this backwards compatible. Or do we want to have any string be a potential mapping?

I concur with "default" and "base_env" even though I'd probably want to be more explicit and use "base_environment".

jaimergp:
> I do like the mapping idea, but we need to make this backwards compatible. Or do we want to have any string be a potential mapping?

We should be able to accept a string (original behaviour) or a mapping of conditions to string. Since we have total control over condition keys, we can either get creative with `target_environment==base` and `target_environment!=base` (but those are not evaluated, just strings we interpret in that way), or using op-less strings like `target_environment_is_base` and `target_environment_is_not_base`.

I don't care much as long as things are documented nicely.

marcoesters:
> > I do like the mapping idea, but we need to make this backwards compatible. Or do we want to have any string be a potential mapping?
> 
> We should be able to accept a string (original behaviour) or a mapping of conditions to string. Since we have total control over condition keys, we can either get creative with `target_environment==base` and `target_environment!=base` (but those are not evaluated, just strings we interpret in that way), or using op-less strings like `target_environment_is_base` and `target_environment_is_not_base`.
> 
> I don't care much as long as things are documented nicely.

I added the `target_environment_is_base`/`_not_base` keys and added a test. This should allow for both the new v2 and old v1 behavior to co-exist.

docs/source/defining-shortcuts.md
```diff
@@ -55,6 +55,42 @@ The full list of available placeholders is available at {ref}`placeholders`.
...
+If environments are supported, different naming schemes can be specified for installations into
+the base environment and non-base environments.
+To do this, the `name` property must be a dictionary with the keys `target_environment_is_base`
+and `target_environment_is_not_base` for installations into the base and non-base environment,
+respectively.
+
+
+The example below creates a shortcut called with the name "Launch Turtle" if installed into the
+base environment. If installed into an environment called, e.g., `turtle`, the name of the shortcut
+is "Lauch Tutrle (turtle)". This was the default behavior of `menuinst` version 1.
```
beeankha, line 69:
```suggestion
is "Launch Turtle (turtle)". This was the default behavior of `menuinst` version 1.
```

menuinst/_schema.py
```diff
@@ -36,8 +36,19 @@ class BasePlatformSpecific(BaseModel):
     * Default value is always ``None``.
     """
 
-    name: Optional[constr(min_length=1)] = None
-    "The name of the menu item"
+    name: Optional[
+        Union[
+            constr(min_length=1),
+            Dict[constr(regex=r"^target_environment_is(_not)?_base$"), constr(min_length=1)],
```
jaimergp, line 42: I think it's easier to just enumerate the two options here with a Literal:

```suggestion
            Dict[Literal["target_environment_is_base", "target_environment_is_not_base"], constr(min_length=1)],
```

menuinst/_schema.py
```diff
@@ -340,8 +351,17 @@ class Platforms(BaseModel):
 class MenuItem(BaseModel):
     "Instructions to create a menu item across operating systems."
 
-    name: constr(min_length=1) = ...
-    "The name of the menu item."
+    name: Union[
+        constr(min_length=1),
+        Dict[constr(regex=r"^target_environment_is_(not_)?base$"), constr(min_length=1)],
```
jaimergp, line 356: Same here. We can even create a TypeVar at the top-level if needed.

menuinst/platforms/base.py
```diff
@@ -37,7 +37,10 @@ def __init__(
         self.prefix = Path(prefix)
         self.base_prefix = Path(base_prefix)
 
-        self.env_name = None
+        if self.prefix.absolute() == self.base_prefix.absolute():
```
jaimergp, line 40: I'm not sure this handles symlinks well. See [discussion here](https://discuss.python.org/t/pathlib-absolute-vs-resolve/2573/19) for more details about `.absolute()` being not recommended in general.

Instead use [`.samefile()`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.samefile) which [does follow symlinks](https://stackoverflow.com/questions/71939549/does-os-path-samefile-follow-symbolic-links).

news/180-support-different-names-outside-base
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Support different name for shortcuts within and outside base environment (support v1 behavior)
```
jaimergp, line 3:
```suggestion
* Support different name for shortcuts within and outside base environment (support v1 behavior). (#180)
```

docs/source/defining-shortcuts.md
```diff
@@ -55,6 +55,42 @@ The full list of available placeholders is available at {ref}`placeholders`.
...
+```json
+{
+  "$schema": "https://json-schema.org/draft-07/schema",
+  "$id": "https://schemas.conda.io/menuinst-1.schema.json",
+  "menu_name": "Python {{ PY_VER }}",
+  "menu_items": [
+    {
+      "name": {
+        "target_environment_is_base": "Launch Turtle",
+        "target_environment_is_not_base": "Launch Turtle ({{ ENV_NAME }})"
```
jaimergp, line 80: I'm thinking of non-conda implementations of the schema (where there's no "base" environment, like pixi or micromamba), and whether they'll have to deal with these semantics. They'll probably ignore `*_is_base`, but will have to handle `*_is_not_base`. 

In a way, it looks to me that `target_environment_is_not_base` is the default case, while `target_environment_is_base` seems to be a special case in the grand scheme of things.

Should we have this as `default` and `target_environment_is_base`, instead? I know I'm bikeshedding a bit, but I wanted to make sure that other angles are considered.

cc @wolfv for awareness.

marcoesters: I think `default` is ambiguous and what's the most common use case for micromamba and pixi is not necessarily the use case for others. I would say that for `conda`, the base environment case is much more common since the console shortcut, e.g., is installed into the base environment via the installer.

I think explicit is better than implicit.

menuinst/_schema.py
marcoesters: I ended up using a `TypedDict`. Using `Literal` as dictionary keys did not get processed well by the schema script (it discarded the key restrictions)

marcoesters: `Literal` does not seem to produce the outcome that's expected because the `schema_json` function drops the `Literal` entirely.

I tried a `TypedDict`, but that will make things more complicated since `typing_extensions` are needed for python 3.8 and the schema fails to validate (even though it looks okay to me).

marcoesters: I think I got it to work now. The way it is now should make it extensible in case we want to add additional naming conditions later.

menuinst/platforms/base.py
```diff
@@ -137,6 +140,14 @@ def __init__(self, menu: Menu, metadata: dict):
...
         self._data = self._initialize_on_defaults(metadata)
         self.metadata = self._flatten_for_platform(self._data)
+        if isinstance(self.metadata["name"], dict):
+            if self.menu.env_name:
+                name = self.metadata["name"].get("target_environment_is_not_base", "")
+            else:
+                name = self.metadata["name"].get("target_environment_is_base", "")
+            if not name:
+                raise KeyError("Cannot parse `name` from dictionary representation.")
+            self.metadata["name"] = name
```
jaimergp, line 150: At this point we have not rendered the `{{ variables }}`, right?

menuinst/platforms/base.py
```diff
@@ -37,7 +37,10 @@ def __init__(
         self.prefix = Path(prefix)
         self.base_prefix = Path(base_prefix)
 
-        self.env_name = None
+        if self.prefix.samefile(self.base_prefix):
+            self.env_name = None
+        else:
+            self.env_name = self.prefix.name
```
jaimergp, line 43: Maybe this is the value we should have in `placeholders()` for `ENV_NAME`?

menuinst/platforms/base.py
marcoesters: That's correct.

The tests for a dictionary representation of the name would have failed by then, so even if code changes render the name at that point, the tests will catch that.

menuinst/platforms/base.py
```diff
@@ -37,7 +37,10 @@ def __init__(
         self.prefix = Path(prefix)
         self.base_prefix = Path(base_prefix)
 
-        self.env_name = None
+        if self.prefix.samefile(self.base_prefix):
+            self.env_name = None
```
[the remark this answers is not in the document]
marcoesters: That's actually quite tricky. "base environment" is a concept, that doesn't mean that the base environment has to be called "base" in all implementations. On the other hand, `self.base_prefix.name` is incorrect because that's the directory name of the installation, not the name of the environment.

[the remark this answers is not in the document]
jaimergp: Since we already have the installation directory name in https://github.com/conda/menuinst/blob/f3c25ed1f9a825836ca3980d4e445db3ff0b0781/menuinst/platforms/base.py#L72, so I’d say we go for `base`.

menuinst/platforms/base.py
```diff
@@ -75,7 +75,7 @@ def placeholders(self) -> Dict[str, str]:
             "DISTRIBUTION_NAME": self.base_prefix.name,
             "BASE_PYTHON": str(self.base_prefix / "bin" / "python"),
             "PREFIX": str(self.prefix),
-            "ENV_NAME": self.prefix.name,
+            "ENV_NAME": self.env_name or "",
```
jaimergp, line 78: Can this be false-y? 🤔 I think we don't need the `or ""` part, right?

marcoesters: Now that I look at it again, I don't think so. I guess I was overly cautious here.

jaimergp (approved):

jaimergp:
This will need a CEP update.

PR #194, opened 2024-04-25
Add support for NSSupportsAutomaticGraphicsSwitching

marcoesters:
### Description

Some MacBooks have two GPUs, a discrete and an integrated GPU. While the former is more powerful, it also consumes more power. By default, the [discrete GPU](https://developer.apple.com/library/archive/qa/qa1734/_index.html) is used.

To allow apps using OpenGL to use the integrated GPU and be more energy-efficient, the [NSSupportsAutomaticGraphicsSwitching](https://developer.apple.com/documentation/bundleresources/information_property_list/nssupportsautomaticgraphicsswitching) needs to be set to `true`.

`menuinst` currently does not support this flag - this PR adds this property to the schema.

### Checklist - did you ...

### DIFF
--- a/menuinst/_schema.py
+++ b/menuinst/_schema.py
@@ -295,6 +295,8 @@ class MacOS(BasePlatformSpecific):
     If true, prevent a universal binary from being run under
     Rosetta emulation on an Intel-based Mac.
     """
+    NSSupportsAutomaticGraphicsSwitching: Optional[bool] = None
+    """If true, allows an OpenGL app to utilize the integrated GPU."""
     UTExportedTypeDeclarations: Optional[List[UTTypeDeclarationModel]] = None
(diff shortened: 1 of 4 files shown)


jaimergp:
Thanks! Can we add a link to the `NS*` keys documentation in line 180?
This will have to be added to the CEP schema, either via a new CEP or a small edit because the change is not too drastic.

marcoesters:
> Thanks! Can we add a link to the `NS*` keys documentation in line 180? This will have to be added to the CEP schema, either via a new CEP or a small edit because the change is not too drastic.

The `NS*` keys do not have a neat list like the `CF*` and `LS*` keys - or at least, I haven't been able to find them. In fact, some `NS*` keys are part of the `CF*`documentation (not this key though).

jaimergp (approved):

PR #197, opened 2024-04-29
Maintenance: bump old actions, add dependabot, fix macOS runners configuration

jaimergp:
### Description

Canaries are failing on macOS due to outdated actions. Fixing and preventing future errors.

### Checklist - did you ...

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -43,7 +43,7 @@ jobs:
             python-version: "3.9"
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           fetch-depth: 0
 
@@ -55,7 +55,7 @@ jobs:
           echo "XDG_CURRENT_DESKTOP=GNOME" >> $GITHUB_ENV
 
       - name: Install miniforge
-        uses: conda-incubator/setup-miniconda@v2
+        uses: conda-incubator/setup-miniconda@v3
         with:
           miniforge-version: latest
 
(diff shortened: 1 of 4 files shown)


.github/workflows/tests.yml
```diff
@@ -43,7 +43,7 @@ jobs:
             python-version: "3.9"
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
```
kenodegard, line 46: use commit hashes instead?

marcoesters (commented):
I'm more familiar with renovate than dependabot, but I have the following questions/suggestions:

* Many actions use only major version numbers. It would be more transparent to add minor and patch numbers, too. In renovate, this also results in better pull request descriptions.
* Will dependabot replace the version numbers with commit hashes eventually? This would result in better security: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

.github/workflows/tests.yml
jaimergp: We are adding dependabot here to handle that automatically.

kenodegard: AFAIK dependabot will replace tags with tags and commits hashes with commits hashes

jaimergp: Oh, really? That's unfortunate 😬 Let's see if there's a tool to automatically convert.

jaimergp: Used https://github.com/mheap/pin-github-action and some post-processing. All pinned now.

marcoesters: Do we want to add `<major>.<minor>.<patch>` version numbers to all of them to make it easier to see which version is used exactly?

jaimergp: Ah, I see why now. Cool. Added!

marcoesters (approved):

PR #200, opened 2024-05-02
Add Microsoft Terminal profile shortcut

### LINKED ISSUE #196: Add Windows Terminal profile option
### Checklist

### What is the idea?

The Windows Terminal app contains profiles for terminal-based applications. These are stored in a `settings.json` file that can be appended. The [Microsoft documentation](https://learn.microsoft.com/en-us/windows/terminal/dynamic-profiles) contains a few examples of dynamic profiles.

### Why is this needed?

Conda packages such as the console shortcuts for Miniconda, Anaconda Distribution, and Miniforge can be opened in the Windows Terminal, but do not appear in the profile list. This would be a great convenience feature to use the tab feature of the Windows Terminal instead of having to open a new Window each time.

### What should happen?

A new option should be added to allow adding Windows Terminal profiles.

### Additional Context

This is currently focused on the Windows Terminal, but the schema should be updated such that it could potentially be expanded to other platforms if the profile config files have plain-text configuration files.

marcoesters:
### Description

The Windows Terminal provides the option to application to its profile. This is a handy feature for terminal-based applications.

This PR adds the option to create a Microsoft Terminal profile. It is limited to non-admin installations because, to my knowledge, there is no global `%LOCALAPPDATA%` file in which the settings file resides.

For now, the terminal profiles are identified by their name, which the terminal uses to create a GUID. It is, in principle, possible to create our own GUIDs using the `uuid` module from other types of heuristics, but I decided to keep using the name of the menu item to avoid having profiles with duplicate names.

Closes #196 

### Checklist - did you ...


menuinst/platforms/win.py
```diff
@@ -64,6 +65,22 @@ def quick_launch_location(self) -> Path:
     def desktop_location(self) -> Path:
         return Path(windows_folder_path(self.mode, False, "desktop"))
 
+    @property
+    def terminal_profile_location(self) -> Path:
+        if self.mode == "system":
+            warnings.warn("Terminal profiles are not available for system level installs")
```
jaimergp, line 71: This is a warning for the end user, right? In that case:

```suggestion
            log.warn("Terminal profiles are not available for system level installs")
```

I think I messed up the other `warnings.warn` calls to 🤔 They should have been `logging.Logger.warning` calls too.

menuinst/platforms/win.py
```diff
@@ -64,6 +65,22 @@ def quick_launch_location(self) -> Path:
...
+            return None
+        profile_path = (
+            Path(windows_folder_path(self.mode, False, "localappdata"))
+            / "Packages"
+            / "Microsoft.WindowsTerminal_8wekyb3d8bbwe"
+            / "LocalState"
+            / "settings.json"
+        )
```
jaimergp, line 79:
```suggestion
        profile_path = Path(
            windows_folder_path(self.mode, False, "localappdata"),
            "Packages",
            "Microsoft.WindowsTerminal_8wekyb3d8bbwe",
            "LocalState",
            "settings.json",
        )
```

Silly thing to save a couple lines, if you like it better.

menuinst/platforms/win.py
```diff
@@ -64,6 +65,22 @@ def quick_launch_location(self) -> Path:
...
 
+    @property
+    def terminal_profile_location(self) -> Path:
+        if self.mode == "system":
+            warnings.warn("Terminal profiles are not available for system level installs")
+            return None
+        profile_path = (
+            Path(windows_folder_path(self.mode, False, "localappdata"))
+            / "Packages"
+            / "Microsoft.WindowsTerminal_8wekyb3d8bbwe"
```
jaimergp, line 76: Is this the actual folder name? That random string thing looks autogenerated but maybe is not?

tests/data/jsons/windows-terminal.json
jaimergp, line 1: Can we format this file as the other data jsons? I think they use 2 spaces too.

menuinst/platforms/win.py
marcoesters: I agree that it looks that way, but it is not auto-generated: https://learn.microsoft.com/en-us/windows/terminal/install#settings-json-file

I should point out in the documentation though that it adds to the stable release. Or I expand the functionality to install into all instances (stable, preview, and unpackaged) it finds. What do you think?

jaimergp: Stable is enough, I'd say. No strong opinion if you want to add more, though.

marcoesters: Might as well

menuinst/platforms/win.py
```diff
@@ -293,6 +335,45 @@ def _process_command(self, with_arg1=False) -> Tuple[str]:
...
+            if self.metadata.get("icon"):
+                profile_data["icon"] = self.render_key("icon")
+            if self.metadata.get("working_dir"):
+                profile_data["startingDirectory"] = self.render_key("working_dir")
+            if index < 0:
+                settings["profiles"]["list"].append(profile_data)
+            else:
+                log.warn(f"Overwriting terminal profile for {name}.")
+                settings["profiles"]["list"][index] = profile_data
+        location.write_text(json.dumps(settings, indent=4))
```
jaimergp, line 375: What's the default indentation level in the factory JSON? I would have expected `2`.

```suggestion
        location.write_text(json.dumps(settings, indent=2))
```

tests/test_api.py
```diff
@@ -294,3 +297,61 @@ def test_url_protocol_association(delete_files):
...
+            "Microsoft",
+            "Windows Terminal",
+            "settings.json",
+        ),
+    ]
+    settings_files = [file for file in settings_files if file.exists()]
+    if not settings_files:
+        pytest.skip("No terminal profile settings file found.")
+    tmpdir = mkdtemp()
+    (Path(tmpdir) / ".nonadmin").touch()
```
jaimergp, line 340:
```suggestion
    (tmp_path / ".nonadmin").touch()
```

Let's use the `tmp_path` fixture instead.

tests/test_api.py
```diff
@@ -294,3 +297,61 @@ def test_url_protocol_association(delete_files):
         url_to_open=url,
         expected_output=url,
     )
+
+
+@pytest.mark.skipif(PLATFORM != "win", reason="Windows only")
+def test_windows_terminal_profiles():
```
jaimergp, line 303:
```suggestion
def test_windows_terminal_profiles(tmp_path):
```

tests/test_api.py
jaimergp: Update instances below too 🙏

tests/test_api.py
```diff
@@ -294,3 +297,61 @@ def test_url_protocol_association(delete_files):
...
+        Path(
+            folder_path("user", False, "localappdata"),
+            "Microsoft",
+            "Windows Terminal",
+            "settings.json",
+        ),
+    ]
+    settings_files = [file for file in settings_files if file.exists()]
+    if not settings_files:
+        pytest.skip("No terminal profile settings file found.")
```
jaimergp, line 338: Is Terminal preinstalled on GHA CI? 🤔 I don't see it in https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md

menuinst/platforms/win.py
marcoesters: It is four spaces without EOL at the end of the file.

tests/test_api.py
```diff
@@ -296,6 +299,63 @@ def test_url_protocol_association(delete_files):
...
+            "settings.json",
+        ),
+        # Unpackaged (Scoop, Chocolatey, etc.)
+        Path(
+            folder_path("user", False, "localappdata"),
+            "Microsoft",
+            "Windows Terminal",
+            "settings.json",
+        ),
+    ]
```
jaimergp, line 335: These are duplicated in `menunist.platforms.win`. I wonder if we should put them in a function that takes `mode` and returns the same list so we can test it here. Thoughts?

marcoesters: I moved it into its own function. I also decided to harden the function in case that seemingly auto-generated string ever gets changed.

tests/test_api.py
marcoesters: Well, that led me down a rabbit hole...

I discovered that Terminal does not have the config files set up after installation. You have to run it to write the files. So, I had to change some of the behavior of how the files are detected. Luckily, writing the a file and opening the terminal does not remove the profiles, so we're good there.

The second problem was that `folder_path` is mocked away in the tests. So, I had to change the tests to create the files inside the mocked directory. Additionally, I had to add a custom mark to use the "true" paths on the CI.

However, I discovered that using chocolatey to install the terminal onto the GitHub runner actually doesn't work even though the installation succeeded: https://github.com/conda/menuinst/actions/runs/9023438085/job/24795241900#step:4:2662

`wt` should have been available, so I have no idea why the CI behaves so differently than a local system. I'm a little stumped here.

tests/conftest.py
```diff
@@ -18,6 +18,12 @@
 PLATFORM = platform_key()
 
 
+def pytest_configure(config):
+    config.addinivalue_line(
+        "markers", "no_mocking_on_ci: Do not mock locations when running on the CI."
```
jaimergp, line 23: The rest of the `conda` repos have this kind of config in `pyproject.toml`. It might be better to follow that trend for consistency.

menuinst/platforms/win_utils/knownfolders.py
```diff
@@ -313,3 +316,35 @@ def folder_path(preferred_mode, check_other_mode, key):
...
+        ],
+        # Unpackaged (Scoop, Chocolatey, etc.)
+        Path(
+            localappdata,
+            "Microsoft",
+            "Windows Terminal",
+            "settings.json",
+        ),
+    ]
+    return profile_locations
```
jaimergp, line 350: Are we not checking for existence here?

tests/conftest.py
marcoesters: Ah, good to know. I wasn't sure since I didn't see any pytest-related parts in `pyproject.toml`. Before I implements this, I'm wondering if we should generally disable mocking on the CI, but that might be out of scope.

menuinst/platforms/win_utils/knownfolders.py
marcoesters: My initial thought was, when the folders are still hard-coded, that these are the "known folders", i.e., where these files could be. However, I can also just check for existence right here, that's true.

In that case, I could also change the tests so that they just create a directory in the tests if no directory is found. That way, it's at least testing something. Does that sound good?

jaimergp: Yep!

marcoesters: Since the terminal doesn't appear to be CI-installable, I decided to do away with all removing the location mocking and just use a mock directory for the terminal test.

I ran into another problem though: on the CI, you are always admin, and the terminal has no definition for an admin profile. So, I had to mock the `user_is_admin` function to return `False` for this test.

tests/conftest.py
```diff
@@ -52,7 +52,7 @@ def tmpdir(tmpdir, request):
 
 
 @pytest.fixture(autouse=True)
-def mock_locations(monkeypatch, tmp_path):
+def mock_locations(monkeypatch, tmp_path, request):
```
jaimergp, line 55: Are you using this?

jaimergp:
```suggestion
def mock_locations(monkeypatch, tmp_path):
```

jaimergp (approved):
LGTM. Feel free to merge once you are ready!

If you have a screenshot of how it looks like on Windows, I think it would be nice to have it posted here. If it's too much trouble, no worries.

marcoesters:
> LGTM. Feel free to merge once you are ready!
> 
> If you have a screenshot of how it looks like on Windows, I think it would be nice to have it posted here. If it's too much trouble, no worries.

I will create one and post it before I merge.

jaimergp (approved):

marcoesters:
Here is a menu file to create the git bash terminal profile and the screenshot shows how it look like in the Windows Terminal.

```json
{
  "$schema": "https://json-schema.org/draft-07/schema",
  "$id": "https://schemas.conda.io/menuinst-1.schema.json",
  "menu_name": "git bash",
  "menu_items": [
    {
      "name": "git bash",
      "command": [
          "C:\\Program Files\\Git\\bin\\bash.exe",
          "-li"
      ],
      "platforms": {
        "win": {
            "quicklaunch": false,
            "desktop": false,
            "windows_terminal_profile": "Git Bash",
            "icon": "C:\\Program Files\\Git\\mingw64\\share\\git\\git-for-windows.ico"
        }
      }
    }
  ]
}
```

Screenshot 2024-05-15 at 1 40 22 PM

PR #201, opened 2024-05-03
Render all menu name instances for Linux

marcoesters:
### Description

Some menu names are not rendered for Linux shortcuts. This PR adds the the remaining instances.

### Checklist - did you ...

### DIFF
--- a/menuinst/platforms/linux.py
+++ b/menuinst/platforms/linux.py
@@ -42,6 +42,10 @@ class LinuxMenu(Menu):
                 os.environ.get("XDG_DATA_HOME", "~/.local/share")
             ).expanduser()
 
+        # The names have to be rendered often, so store them for convenience
+        self._rendered_name = self.render(self.name)
+        self._slugified_name = self.render(self.name, slug=True)
+
         # XML Config paths
         self.system_menu_config_location = (
             self._system_config_directory / "menus" / "applications.menu"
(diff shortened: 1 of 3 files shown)


menuinst/platforms/linux.py
```diff
@@ -42,16 +42,18 @@ def __init__(self, *args, **kwargs):
                 os.environ.get("XDG_DATA_HOME", "~/.local/share")
             ).expanduser()
 
+        # The names have to be rendered often, so store them for convenience
+        self._rendered_name = self.render(self.name)
+        self._slugified_name = self.render(self.name, slug=True)
+
```
jaimergp, line 48: Should we memoize `render` instead?

marcoesters: The main problem I was trying to solve here is the human error component where rendering was forgotten. I could make this a cached property though if that makes more sense.

jaimergp: I think folks would also forget to look for the private attribute. What if we make .name always rendered and offer a raw_name instead?

marcoesters: I think that would just open up a different can of worms since we need to decide on when to render (and that may clash with #180).

I'm happy to remove the private class members and return the function calls (and even cache the rendering if that's better) if you think that's better.

jaimergp: Ok, let's do that then. Use `render()` explicitly + lrucache the method.

jaimergp (approved):

PR #205, opened 2024-05-07
Improve errors and warnings when shortcuts already exist

### LINKED ISSUE #203: Installation of MacOS apps fail when app already exists
### Checklist

- [X] I searched open reports and couldn't find a duplicate

### What happened?

On Windows and Linux, shortcuts are overwritten when they already exist. On MacOS, `menuinst` exits with an error because it cannot create the `Resources` directory inside the app. This is due to the following lines: https://github.com/conda/menuinst/blob/main/menuinst/platforms/osx.py#L107-L108

```
for path in paths:
    path.mkdir(parents=True, exist_ok=False)
```

The MacOS app should be removed before installing a new one to make the behavior consistent.

### Conda Info

_No response_

### Conda Config

_No response_

### Conda list

_No response_

### Additional Context

_No response_

marcoesters:
### Description

On Windows and Linux, shortcuts are overwritten when they already exist. On MacOS, menuinst exits with an error because it cannot create the Resources directory inside the app.

~Since an app is not a simple file, removing the entire directory before creating the new app is the easiest way to ensure that a proper overwrite was done. The test I fails with the current menuinst.~

I also decided to add warnings to when shortcuts are overwritten. This behavior has always been a little hidden and could cause some surprises if users are not aware of this behavior.

EDIT: after some discussion, a more prudent way is to improve the error message. Individual packages can still remove the app using pre-link features if they want to overwrite.

Closes #203.

### Checklist - did you ...

### DIFF
--- a/docs/source/defining-shortcuts.md
+++ b/docs/source/defining-shortcuts.md
@@ -19,8 +19,9 @@ four keys
         - `win`- see {class}`Windows schema <menuinst._schema.Windows>`
 
 ```{warning}
-`menuinst` will overwrite existing shortcuts, so `menu_name` and `name` must be chosen accordingly.
-Note that MacOS does not use `menu_name`.
+`menuinst` will overwrite existing shortcuts on Linux and Windows, so `menu_name` and `name` must be chosen accordingly.
+
+MacOS apps, however, will not be overwritten. Instead, `menuinst` will abort when an app with the same name exists.
 ```
 
 ```{seealso}
--- a/docs/source/defining-shortcuts.md
+++ b/docs/source/defining-shortcuts.md
@@ -18,6 +18,11 @@ four keys
         - `osx`- see {class}`MacOS schema <menuinst._schema.MacOS>`
         - `win`- see {class}`Windows schema <menuinst._schema.Windows>`
 
+```{warning}
+`menuinst` will overwrite existing shortcuts, so `menu_name` and `name` must be chosen accordingly.
+Note that MacOS does not use `menu_name`.
+```
+
 ```{seealso}
 If you want to learn more, check the {doc}`reference` for full details on the available fields and settings for each platform.
 The JSON configurations follow a well-defined schema documented at {ref}`schema`.
--- a/menuinst/platforms/linux.py
+++ b/menuinst/platforms/linux.py
@@ -231,7 +231,7 @@ class LinuxMenuItem(MenuItem):
 
     def _write_desktop_file(self):
         if self.location.exists():
-            log.warn(f"Overwriting existing file at {self.location}.")
+            log.warning(f"Overwriting existing file at {self.location}.")
 
         lines = [
             "[Desktop Entry]",
--- a/menuinst/platforms/linux.py
+++ b/menuinst/platforms/linux.py
@@ -233,7 +233,7 @@ class LinuxMenuItem(MenuItem):
 
     def _write_desktop_file(self):
         if self.location.exists():
-            log.warning(f"Overwriting existing file at {self.location}.")
+            log.warning("Overwriting existing file at %s.", self.location)
 
         lines = [
             "[Desktop Entry]",
--- a/menuinst/platforms/linux.py
+++ b/menuinst/platforms/linux.py
@@ -230,6 +230,9 @@ class LinuxMenuItem(MenuItem):
         return "bash -c " + shlex.quote(" && ".join(parts))
 
     def _write_desktop_file(self):
+        if self.location.exists():
+            log.warn(f"Overwriting existing file at {self.location}.")
+
         lines = [
             "[Desktop Entry]",
             "Type=Application",
--- a/menuinst/platforms/osx.py
+++ b/menuinst/platforms/osx.py
@@ -76,7 +76,7 @@ class MacOSMenuItem(MenuItem):
     def create(self) -> Tuple[Path]:
         log.debug("Creating %s", self.location)
         if self.location.exists():
-            log.warn(f"Overwriting existing application at {self.location}.")
+            log.warning(f"Overwriting existing application at {self.location}.")
             self.remove()
         self._create_application_tree()
         self._precreate()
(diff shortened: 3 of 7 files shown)


jaimergp:
I wonder now if we should rename existing shortcuts so there's a backup in case we overstepped accidentally.

tests/test_api.py
```diff
@@ -144,6 +144,19 @@ def test_install_prefix(delete_files):
     check_output_from_shortcut(delete_files, "sys-prefix.json", expected_output=sys.prefix)
 
 
+def test_overwrite_existing_shortcuts(delete_files):
```
jaimergp, line 147: Can we try to catch the warnings and ensure they are being produced?

jaimergp:
> I wonder now if we should rename existing shortcuts so there's a backup in case we overstepped accidentally.



Any thoughts about this question?

marcoesters:
> I was thinking about how the macos case is a bit different. On Windows and Unix, the "shortcut file" is just a file whose sole purpose is metadata to launch a program. However in macOS it is a directory, and we might be very unlucky if someone publishes a shortcut named `Finder` (or any other preexisting app). We are writing to `~/Applications` for this reason too, but if we change that behavior in the future this might be an oversight.

This is not a risk for system-critical applications because they are in `/System/Applications`, but your point is well taken. Safari does get installed into `/Applications` though. We do install into `/Applications` as root, too.

> I think that at the very least we should send it to Trash (via [osascript](https://apple.stackexchange.com/questions/50844/how-to-move-files-to-trash-from-command-line) maybe?), instead of removing it. At least they'll have the opportunity for recovery. The annoyance of having multiple copies is not as worse as deleting an important directory.

The problem isn't just that they don't exist, but that `menuinst` will never remove them as far as I understand it. Let's assume we rename them somehow and we have `App.app` and `App (1).app`. Removing the app would leave `App (1).app` behind.

I see the following ways out:

1. Move into Trash, as you said, but with a warning
2. Use subdirectories to isolate the application the way Windows does it - that may be an anti-pattern for MacOS though
3. Crash as before, but with a very descriptive error message (i.e., "remove the existing application or run as with the `--no-shortcuts`" option).

jaimergp:
None is ideal... I'm leaning towards (3), but we should check something first. Does the name of the .app directory affect the display name in the UI? Because it does not (and it gets it from the Plist config inside), maybe we can just add some timestamp to the filename to avoid collisions.

marcoesters:
> None is ideal... I'm leaning towards (3), but we should check something first. Does the name of the .app directory affect the display name in the UI? Because it does not (and it gets it from the Plist config inside), maybe we can just add some timestamp to the filename to avoid collisions.

I generally wouldn't want timestamps because they don't tell the user anything. However, the point is moot since there is a display change: the name next to the Apple symbol in the taskbar shows the name of the file, not the name inside `Info.plist`.

> Wrt to (2), I'm not sure it's an antipattern. I do see a "Chrome apps" subdirectory in my ~/Applications. See https://apple.stackexchange.com/questions/182626/chrome-keeps-creating-application-folder-in-home-folder for some context. Reading there I also learnt that Chrome puts the same .app directories under ~/Library/Application Support/Google/Chrome/Default/Web Applications/LONG_RANDOM_IDENTIFIER/ too, so we could also do that if needed (with a symlink?). Something like ~/Library/Application Support/Menuinst/X/Y/Z.

While it may not be an anti-pattern, it will at least be a breaking change for other developers who have already pushed MacOS apps using `menuinst`. The sym-link option is intriguing though.

For now, a better error message may indeed be the solution.

menuinst/platforms/osx.py
```diff
@@ -74,6 +74,14 @@ def _precreate(self):
             os.symlink(self.render(src), rendered_dest)
 
     def create(self) -> Tuple[Path]:
+        if self.location.exists():
+            message = (
+                f"App already exista at {self.location}. "
```
jaimergp, line 79:
```suggestion
                f"App already exista ats {self.location}. "
```

menuinst/platforms/osx.py
```diff
@@ -74,6 +74,14 @@ def _precreate(self):
...
+                "Please remove the existing shortcut before installing. "
+                "If you used conda to install this package, "
+                "re-run the installation with --no-shortcuts."
```
jaimergp, line 82: I think conda will not raise this exception, and just warn about it. So the message could be instead something like:

```suggestion
                "If you used conda to install this package, "
                "reinstall the package with --force-reinstall to "
                "create the shortcut once the location is cleared."
```

menuinst/platforms/win.py
```diff
@@ -154,6 +154,8 @@ def create(self) -> Tuple[Path, ...]:
             # Notice args must be passed as positional, no keywords allowed!
             # winshortcut.create_shortcut(path, description, filename, arguments="",
             #                             workdir=None, iconpath=None, iconindex=0, app_id="")
+            if Path(path).exists():
+                log.warning(f"Overwriting existing link at {path}.")
```
jaimergp, line 158:
```suggestion
                log.warning("Overwriting existing link at %s.", path)
```

docs/source/defining-shortcuts.md
```diff
@@ -18,6 +18,11 @@ four keys
         - `osx`- see {class}`MacOS schema <menuinst._schema.MacOS>`
         - `win`- see {class}`Windows schema <menuinst._schema.Windows>`
 
+```{warning}
+`menuinst` will overwrite existing shortcuts, so `menu_name` and `name` must be chosen accordingly.
```
jaimergp, line 22: I think we need to update this to cover the macOS logic.

menuinst/platforms/linux.py
```diff
@@ -232,6 +232,9 @@ def _command(self) -> str:
         return "bash -c " + shlex.quote(" && ".join(parts))
 
     def _write_desktop_file(self):
+        if self.location.exists():
+            log.warning(f"Overwriting existing file at {self.location}.")
```
jaimergp, line 236:
```suggestion
            log.warning("Overwriting existing file at %s.", self.location)
```

jaimergp (approved):

PR #212, opened 2024-05-16
Expand variables for working directory and adjust Windows default path

### LINKED ISSUE #211: %HOMEPATH% does not contain the drive name
### Checklist

- [X] I searched open reports and couldn't find a duplicate

### What happened?

In the absence of the `working_dir` key, on Windows, `menuinst` uses `%HOMEPATH%` as the default. However, I have stumbled over this on a system, where the home was not on the `C:` drive. In order to also get the drive, one would need to use `%HOMEDRIVE%`. When googling to find a solution, I also stumbled over a number of reports that these two variables may not be set in some circumstances.

I would therefore suggest to maybe use `%USERPROFILE%` for that purpose. Apparently, it is always set to a full path.

### Conda Info

_No response_

### Conda Config

_No response_

### Conda list

_No response_

### Additional Context

_No response_

marcoesters:
### Description

This PR fixes two bugs:

1. Environment variables are permitted for `working_dir`, but are not expanded. If `working_dir` is set to, e.g, `%USERPROFILE%\workdir` on Windows, `menuinst` would create a literal `%USERPROFILE%\workdir` path directory instead of `%SYSTEMDRIVE%\Users\%USERNAME%\workdir`. Expand these variables to create the correct directory name.
2. The bug reported by #211. `%HOMEPATH%` alone does not contain a drive, which can fail if `%HOMEDRIVE%` is not `C:`. This PR uses a default directory that contains the drive name.

The second item is up for debate. The `menuinst` documentation describes that `working_dir` defaults to the user's home directory. On Windows, there can be two different interpretations: `%USERPROFILE%` or `%HOMEDRIVE%%HOMEPATH%`.

I decided to go with a precedent: [os.path.expanduser](https://docs.python.org/3/library/os.path.html#os.path.expanduser) checks if `%USERPROFILE%` is set and goes with `%HOMEDRIVE%%HOMEPATH%` otherwise. `pathlib.Path.home()` uses the same implementation, so I think `menuinst` should use the same definition as the Python ecosystem.

Instead of using `os.path.expanduser("~")`, I am using the names of the variables though to make the shortcuts be resolved at runtime instead of hard-coding the directories.

Closes #211 

### Checklist - did you ...

### DIFF
--- a/news/212-improve-working-dir-handling
+++ b/news/212-improve-working-dir-handling
@@ -4,7 +4,7 @@
 
 ### Bug fixes
 
-* Expand variables when `creating working_dir` and use `os.path.expanduser` logic for default `working_dir` on Windows. (#211 via #212)
+* Expand variables when creating `working_dir` and use `os.path.expanduser` logic for default `working_dir` on Windows. (#211 via #212)
 
 ### Deprecations
 
--- a/news/212-improve-working-dir-handling
+++ b/news/212-improve-working-dir-handling
@@ -1,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Expand variables when `creating working_dir` and use `os.path.expanduser` logic for default `working_dir` on Windows. (#211 via #212)
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* <news item>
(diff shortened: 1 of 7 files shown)


news/212-improve-working-dir-handling
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Expand variables when `creating working_dir` and use `os.path.expanduser` logic for default `working_dir` on Windows. (#211 via #212)
```
jaimergp, line 7:
```suggestion
* Expand variables when creating `working_dir` and use `os.path.expanduser` logic for default `working_dir` on Windows. (#211 via #212)
```

jaimergp (approved):

PR #218, opened 2024-05-28
Only delete `Menu` when empty

marcoesters:
### Description

The `Menu` directories are currently removed even if multiple apps share the same folder. In other words, if two apps install into the same Start Menu directory, uninstalling one of these apps will remove the entire directory. This is a regression from `menuinst v1`.

The same behavior happens on Linux, which doesn't appear to be the intention since `_remove_this_menu()` is only called if there are no additional items in the `Menu`. The unlinking happens regardless though.

This PR ensures that the `Menu` directory is only removed if the directory associated with `Menu` are indeed empty after all `MenuItems` have been removed.

I also noticed that when removing a Windows Terminal Profile via `conda`, the message "Could not find terminal profile for {name}.", presumably because the `remove` operation is executed twice. I think that message isn't helpful in either case, so I removed it here (not worth a separate PR).

### Checklist - did you ...

### DIFF
--- a/menuinst/platforms/win.py
+++ b/menuinst/platforms/win.py
@@ -36,8 +36,12 @@ class WindowsMenu(Menu):
         return (self.start_menu_location,)
 
     def remove(self) -> Tuple[os.PathLike]:
-        log.debug("Removing %s", self.start_menu_location)
-        shutil.rmtree(self.start_menu_location, ignore_errors=True)
+        # Only remove if the Start Menu directory is empty in case applications share a folder.
+        try:
+            next(Path(self.start_menu_location).iterdir())
+        except StopIteration:
+            log.debug("Removing %s", self.start_menu_location)
+            shutil.rmtree(self.start_menu_location, ignore_errors=True)
         return (self.start_menu_location,)
 
     @property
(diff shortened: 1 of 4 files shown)


menuinst/platforms/win.py
```diff
@@ -36,8 +36,14 @@ def create(self) -> Tuple[os.PathLike]:
...
-        log.debug("Removing %s", self.start_menu_location)
-        shutil.rmtree(self.start_menu_location, ignore_errors=True)
+        # Only remove if the Start Menu directory is empty in case applications share a folder.
+        try:
+            menu_location = Path(self.start_menu_location)
+            if menu_location.exists():
+                next(menu_location.iterdir())
+        except StopIteration:
+            log.debug("Removing %s", self.start_menu_location)
+            shutil.rmtree(self.start_menu_location, ignore_errors=True)
```
jaimergp, line 46:
```suggestion
        menu_location = Path(self.start_menu_location)
        if menu_location.exists():
            try:
                # Check directory contents. If empty, it will raise StopIteration
                # and only in that case we delete the directory.
                next(menu_location.iterdir())
            except StopIteration:
                log.debug("Removing %s", self.start_menu_location)
                shutil.rmtree(self.start_menu_location, ignore_errors=True)
```

This reads a bit clearer to me. Let me know what you think.

jaimergp (approved):

PR #263, opened 2024-10-15
Drop Python 3.8 from CI test matrix.

marcoesters:
### Description

`hypothesis`, which is installed to run tests, has dropped Python 3.8 support, which leads to test failures. Drop Python 3.8 from the test matrix and bump the MacOS minimum Python version to 3.9.

Additionally, run `apt-get update` before running `apt-get install` so that `apt` finds all required packages on Ubuntu.

### Checklist - did you ...

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -33,14 +33,12 @@ jobs:
       fail-fast: false
       matrix:
         os: [windows-latest, ubuntu-latest, macos-13]
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
         exclude:
           - os: macos-13
             python-version: "3.11"
           - os: macos-13
             python-version: "3.10"
-          - os: macos-13
-            python-version: "3.9"
 
     steps:
       - uses: actions/checkout@d632683dd7b4114ad314bca15554477dd762a938 #v4.2.0
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -48,6 +48,7 @@ jobs:
       - name: Add Linux dependencies
         if: startsWith(matrix.os, 'ubuntu')
         run: |
+          sudo apt-get update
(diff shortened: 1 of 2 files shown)


jaimergp:
Do we need to update the Python requirements in pyproject.toml and the conda recipe? This is a dependency of conda, and conda-libmamba-solver has already dropped 3.8 in the upcoming v2 compatibility PR, so it's a matter of time anyway...

jaimergp:
xref https://github.com/conda/conda/issues/14201, for context.

marcoesters:
> Do we need to update the Python requirements in pyproject.toml and the conda recipe? 

Not immediately because the code is still Python 3.8 compatible. I would follow the guidance in the cross-referenced issue. Since it is a dependency of `conda`, we shouldn't be dropping support before they do.

jaimergp:
Note we won't be able to provide 3.8 packages in conda-forge for the next release because it was dropped from the CI matrix.

marcoesters:
> Note we won't be able to provide 3.8 packages in conda-forge for the next release because it was dropped from the CI matrix.

I think those are separate, albeit related, concerns and beyond the scope of this PR. It looks like the ecosystem as a whole need to come up with a deprecation policy.

larsoner (approved):
LGTM +1 for merge

larsoner:
... in the meantime this fixes broken CIs which is helpful so I'll merge, even if we might want to do more in follow-ups regarding properly dropping in `pyproject.toml` and elsewhere. Thanks @marcoesters !

jaimergp:
Thank you @larsoner for pushing this. Just for context (in this case it was ok to merge), we have been usually leaving the final merge decision to the PR author when they are a contributor already. So for the future I think it's fair to "approve and wait", and if there's something blocked by the PR itself, kindly point it out for awareness.

marcoesters:
And since @jaimergp had questions/objections, I would have waited for his approval here. I generally want to make sure that they are all addressed.

However since he said it was okay to merge, all is well that ends well.

larsoner:
Okay makes sense 👍

PR #265, opened 2024-10-17
Add SingleMainWindow support

### LINKED ISSUE #264: Add SingleMainWindow support for Linux items
### Checklist

### What is the idea?

To set the icon of Qt applications properly on latest Ubuntu I need to create the following `.desktop` file:
```
[Desktop Entry]
Type=Application
Version=1.5
Name=MNE-Python
StartupWMClass=MNE-Python
Icon=/home/larsoner/python/mne-python/mne/icons/mne_default_icon.png
SingleMainWindow=true
NoDisplay=true
```
This isn't meant to be a launcher, just create an association between the given Qt window's `StartupWMClass` and an `Icon`. There doesn't seem to be any runtime Qt method to get the icon to change, so without this `.desktop` file we're stuck with the default gear icon (ignore the Signal icon above it):

Image

All the options are there *except* the `SingleMainWindow` one, which is helpful because it takes this default behavior of having a "New Window" prompt:

Image

And allows it to be removed, which is what I want:

Image

### Why is this needed?

Anyone who wants to remove the "New Window" option on latest Ubuntu will benefit from this feature.

### What should happen?

☝ 

### Additional Context

N/A

larsoner:
Closes #264, see description there.

Probably minor enough not to need a doc or test update? I just did `git grep NoDisplay` and added stuff in alphabetical order in the three places it showed up, hopefully that's good enough :crossed_fingers:

### DIFF
--- a/menuinst/_schema.py
+++ b/menuinst/_schema.py
@@ -173,8 +173,7 @@ class Linux(BasePlatformSpecific):
     "Hint that the app prefers to be run on a more powerful discrete GPU if available."
     SingleMainWindow: Optional[bool] = None
     """
-    Hint to avoid offering a UI to launch another window of the app.
-    This is useful for removing the 'New Window' option in the app's context menu.
+    Do not show the 'New Window' option in the app's context menu.
     """
     StartupNotify: Optional[bool] = None
     """
--- a/menuinst/_schema.py
+++ b/menuinst/_schema.py
@@ -171,6 +171,11 @@ class Linux(BasePlatformSpecific):
     """
     PrefersNonDefaultGPU: Optional[bool] = None
     "Hint that the app prefers to be run on a more powerful discrete GPU if available."
+    SingleMainWindow: Optional[bool] = None
+    """
+    Hint to avoid offering a UI to launch another window of the app.
+    This is useful for removing the 'New Window' option in the app's context menu.
+    """
     StartupNotify: Optional[bool] = None
     """
     Advanced. See `Startup Notification spec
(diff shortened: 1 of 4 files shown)


menuinst/_schema.py
```diff
@@ -171,6 +171,11 @@ class Linux(BasePlatformSpecific):
     """
     PrefersNonDefaultGPU: Optional[bool] = None
     "Hint that the app prefers to be run on a more powerful discrete GPU if available."
+    SingleMainWindow: Optional[bool] = None
+    """
+    Hint to avoid offering a UI to launch another window of the app.
+    This is useful for removing the 'New Window' option in the app's context menu.
```
marcoesters, line 177:
```suggestion
    Do not show the 'New Window' option in the app's context menu.
```

This seems more direct. Doesn't have to be adopted exactly like that, but I think the way it's written now is a little difficult to understand.

marcoesters (changes requested):
This PR needs a news file so that we can add this to the release notes easily. You can check out the `news` directory for the template and an example.

larsoner:
Thanks for the quick review, done @marcoesters !

news/265-add-singlemainwindow-support
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Add support for `SingleMainWindow` in the Linux platform-specific options to allow disabling the `New Window` option in the app's context menu.
```
marcoesters, line 3:
```suggestion
* Add support for `SingleMainWindow` in the Linux platform-specific options to allow disabling the `New Window` option in the app's context menu. (#264 via #265)
```

marcoesters:
> Thanks for the quick review, done @marcoesters !

Thanks for adding the file! I will approve as soon as your CLA goes through.

jaimergp:
@marcoesters is the CLA still unreviewed? 🤔

jaimergp (approved):

PR #281, opened 2024-11-24
Dump schema descriptions

jaimergp:
### Description

We were not dumping this in the JSON document because pydantic doesn't really support the Sphinx syntax. Use `Field` directly. The diff is big but the changes can be seen in the `.json` file instead of the `.py`, for simplicity.

### Checklist - did you ...

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -13,6 +13,7 @@ dependencies:
   - cffi=1.15.1
   - charset-normalizer=2.1.1
   - colorama=0.4.6
+  - commonmark=0.9.1
   - cryptography=38.0.3
   - docutils=0.19
   - idna=3.4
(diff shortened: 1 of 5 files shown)


marcoesters (commented):
The descriptions contain a lot of line feed characters that don't seem to be necessary. Also, do we want to use sphinx-style links?

jaimergp:
> Also, do we want to use sphinx-style links?

I'd rather not, but `autodoc_sphinx` expects RST syntax in the descriptions, I'm afraid :( Unfortunately, RST is not rendered properly in IDE autocompletion (at least in VS Code). Let me see if there's a way to have autodoc_sphinx take MD.

jaimergp:
Well, terribly hacky but it works! Let me know what you think.

marcoesters (approved):

PR #283, opened 2024-11-25
Use `$schema` and `$id` in the right way

### LINKED ISSUE #282: `$schema` and `$id` are being used in the wrong way
### Checklist

### What happened?

I think I got this wrong when I first designed the schema.

- The menuinst JSON schema ITSELF should define a `$schema` key that points to the JSON Schema dialect.
- Each shortcut should then set their `$schema` key so it points to the menuinst JSON schema corresponding to that version!

With examples:

Each shortcut is doing this weird thing:

```json
{
    "$schema": "https://json-schema.org/draft-07/schema",
    "$id": "https://schemas.conda.io/menuinst-1.schema.json",
    "menu_name": "Example",
    ...
}
```

But a shortcut should look like this:

```json
{
    "$schema": "https://schemas.conda.io/menuinst-1.schema.json",
    "menu_name": "Example",
    ...
}
```

Then `https://schemas.conda.io/menuinst-1.schema.json` itself should have this in the top-level:

```json
{
  "$schema": "https://json-schema.org/draft-07/schema",
  "title": "MenuInstSchema",
  "description": "Metadata required to create menu items across operating systems with ``menuinst``.",
  "type": "object",
  "properties": {
    "$id": {
      "title": "$Id",
      "description": "Version of the menuinst schema.",
      "enum": [
        "https://schemas.conda.io/menuinst-1.schema.json"
      ],
      "type": "string"
    },
    "$schema": {
      "title": "$Schema",
      "description": "Standard of the JSON schema we adhere to.",
      "enum": [
        "https://json-schema.org/draft-07/schema"
      ],
      "type": "string"
    },
  ...
}
```

### Conda Info

```shell

```

### Conda Config

```shell

```

### Conda list

```shell

```

### Additional Context

_No response_

jaimergp:
### Description

Closes #282 

### Checklist - did you ...

### DIFF
--- a/menuinst/data/menuinst-1.schema.json
+++ b/menuinst/data/menuinst-1.schema.json
@@ -8,7 +8,7 @@
       "description": "DEPRECATED. Use ``$schema``.",
       "default": "https://schemas.conda.org/menuinst-1.schema.json",
       "deprecated": true,
-      "markdownDescription": "DEPRECATED. Use `$schema`.",
+      "markdownDescription": "DEPRECATED. Use ``$schema``.",
       "minLength": 1,
       "type": "string"
     },
--- a/menuinst/data/menuinst-1.schema.json
+++ b/menuinst/data/menuinst-1.schema.json
@@ -8,7 +8,7 @@
       "description": "DEPRECATED. Use ``$schema``.",
       "default": "https://schemas.conda.org/menuinst-1.schema.json",
       "deprecated": true,
-      "markdownDescription": "DEPRECATED. Use ``$schema``.",
+      "markdownDescription": "DEPRECATED. Use `$schema`.",
       "minLength": 1,
       "type": "string"
     },
--- a/menuinst/data/menuinst-1.schema.json
+++ b/menuinst/data/menuinst-1.schema.json
@@ -1,6 +1,6 @@
 {
   "title": "MenuInstSchema",
-  "description": "Metadata required to create menu items across operating systems with ``menuinst``.",
+  "description": "Metadata required to create menu items across operating systems with `menuinst`.",
   "type": "object",
   "properties": {
     "$id": {
@@ -8,6 +8,7 @@
       "description": "DEPRECATED. Use ``$schema``.",
       "default": "https://schemas.conda.org/menuinst-1.schema.json",
       "deprecated": true,
+      "markdownDescription": "DEPRECATED. Use ``$schema``.",
       "minLength": 1,
       "type": "string"
     },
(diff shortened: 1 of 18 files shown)


jaimergp:
xref https://github.com/pydantic/pydantic/issues/1478

jaimergp:
I'm not sure about the migration path for this change. Need to take a look at the validation code in conda-build, but with this change it looks like we could just use JSON schema directly without having `menuinst` around.

We should also publish the schema to jsonstore.

Questions to address:

- How do we version the schema? I've seen several options:
  - Independent versioning with SchemaVer or regular SemVer. Right now we just have _the_ schema, but it will keep evolving as we are seeing in https://github.com/conda/ceps/pull/98.
  - Same as menuinst? I don't like this, should be tied to a tool agnostic method.
- How do we name the schema filename once we decide on a version scheme?
- Do we keep all schema versions around in the repo?
- Think of runtime validation.

marcoesters (approved):

marcoesters:
> I'm not sure about the migration path for this change. Need to take a look at the validation code in conda-build, but with this change it looks like we could just use JSON schema directly without having `menuinst` around.
> 
> We should also publish the schema to jsonstore.
> 
> Questions to address:
> 
>     * How do we version the schema? I've seen several options:
>       
>       * Independent versioning with SchemaVer or regular SemVer. Right now we just have _the_ schema, but it will keep evolving as we are seeing in [CEP-11: Update schema with menuinst 2.1.x and 2.2.x changes ceps#98](https://github.com/conda/ceps/pull/98).
>       * Same as menuinst? I don't like this, should be tied to a tool agnostic method.
> 
>     * How do we name the schema filename once we decide on a version scheme?
> 
>     * Do we keep all schema versions around in the repo?
> 
>     * Think of runtime validation.



> I'm not sure about the migration path for this change. Need to take a look at the validation code in conda-build, but with this change it looks like we could just use JSON schema directly without having `menuinst` around.
> 
> We should also publish the schema to jsonstore.
> 
> Questions to address:
> 
>     * How do we version the schema? I've seen several options:
>       
>       * Independent versioning with SchemaVer or regular SemVer. Right now we just have _the_ schema, but it will keep evolving as we are seeing in [CEP-11: Update schema with menuinst 2.1.x and 2.2.x changes ceps#98](https://github.com/conda/ceps/pull/98).
>       * Same as menuinst? I don't like this, should be tied to a tool agnostic method.
> 
>     * How do we name the schema filename once we decide on a version scheme?
> 
>     * Do we keep all schema versions around in the repo?
> 
>     * Think of runtime validation.

This should probably be an issue to discuss this more, but I think schema version and `menuinst` version should be independent. The major versions will probably coincide though.

I don't know much about `SchemaVer`, but if this is commonly used, I don't see why we wouldn't use that one for the schema. For the file name, I would then prefer something like `menuinst.schema.<SchemaVer>.json` since `SchemaVer` uses dashes.

jaimergp:
@wolfv, any thoughts about this?

wolfv:
I just removed `$id` from parsing in the rust version, too.

menuinst/data/menuinst-1.schema.json
```diff
@@ -1,30 +1,36 @@
...
-  "description": "Metadata required to create menu items across operating systems with ``menuinst``.",
+  "description": "Metadata required to create menu items across operating systems with `menuinst`.",
   "type": "object",
   "properties": {
     "$id": {
       "title": "$Id",
       "description": "DEPRECATED. Use ``$schema``.",
       "default": "https://schemas.conda.org/menuinst-1.schema.json",
       "deprecated": true,
+      "markdownDescription": "DEPRECATED. Use ``$schema``.",
```
marcoesters, line 11:
```suggestion
      "markdownDescription": "DEPRECATED. Use `$schema`.",
```

Should it be single back tics like in the other instances?

marcoesters (approved):

PR #288, opened 2024-12-14
Adopt schemaver

### LINKED ISSUE #287: Transition from `$id` + `$schema` to just `$schema`, and versioning
Comes from #282 and #283. 

We need to think how to transition existing schemas and make them pass validation. Maybe this is already the case, but needs to be assessed. There might not be too many published packages with menuinst v2 compat out there anyway.

This also brings the topic of versioning the schema via `$schema`, backwards compatibility, etc. `menuninst` will probably need to stay compatible with all previous schemas indefinitely.

Which versioning scheme should we use? Could be [schemaver](https://docs.snowplow.io/docs/api-reference/iglu/common-architecture/schemaver/), or just good old semver.

Thoughts @conda/constructor @wolfv?

jaimergp:
### Description

Closes #287. Paired with https://github.com/conda/conda-build/pull/5569.

### Checklist - did you ...

- [x] Add / update outdated documentation?


menuinst/_schema.py
```diff
@@ -27,7 +27,8 @@
 
 log = getLogger(__name__)
 SCHEMA_DIALECT = "http://json-schema.org/draft-07/schema#"
-SCHEMA_VERSION = "1"
+# We follow schemaver
+SCHEMA_VERSION = "1-1-0"
```
marcoesters, line 31: Should we always use the latest version here, similar to what you're doing in `docs/source/conf.py`?

menuinst/platforms/base.py
```diff
@@ -22,6 +22,7 @@
 )
 
 log = getLogger(__name__)
+SCHEMA_VERSION = "1-1-0"
```
marcoesters, line 25: Same here, should we detect the latest version here instead of hard-coding it in two places?

jaimergp: I don't think we should here, because it's a runtime bit of the package.

menuinst/_schema.py
jaimergp: We'd run into a catch-22, because to bump it we'd need to first copy the latest file with a new name, and that feels clunky and hard to track via diffs and code reviews.

We could import from `menuinst.platforms.base` though, but I didn't do that because (1) we would start depending on the project itself, and so far this module was standalone, and (2) this feels like the natural place to define the schema version. That's why I "tied" these hardcoded values via a test instead.

What do you think? Would you consider this a blocker?

marcoesters (approved):
Sorry about that! That one got lost in the holiday shuffle.

PR #310, opened 2025-03-04
Replace deprecated test functions

marcoesters:
### Description

`conda.testing.integration.run_command` has been deprecated. Use `conda.testing.fixtures.conda_cli` instead.

### Checklist - did you ...

### DIFF
--- a/news/310-replace-deprecated-run-command
+++ b/news/310-replace-deprecated-run-command
@@ -1,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* <news item>
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* Replace deprecated `run_command` function with `conda_cli`. (#310)
--- a/tests/test_conda.py
+++ b/tests/test_conda.py
@@ -12,7 +12,7 @@ from tempfile import NamedTemporaryFile
 
 import pytest
 from conda.models.version import VersionOrder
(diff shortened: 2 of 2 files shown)


kenodegard (changes requested):
Several observations:
1. `conda_cli` is a pytest fixture and thus cannot be called directly
2. While we can import it as you've done here to expose the fixture to all tests in this file it is preferred to use the `pytest_plugins` constant to keep linters happy https://docs.pytest.org/en/stable/how-to/writing_plugins.html#requiring-loading-plugins-in-a-test-module-or-conftest-file
3. I suggest replacing `new_environment` with the `tmp_env` fixture instead

marcoesters:
> Several observations:
> 
>     1. `conda_cli` is a pytest fixture and thus cannot be called directly
> 
>     2. While we can import it as you've done here to expose the fixture to all tests in this file it is preferred to use the `pytest_plugins` constant to keep linters happy https://docs.pytest.org/en/stable/how-to/writing_plugins.html#requiring-loading-plugins-in-a-test-module-or-conftest-file
> 
>     3. I suggest replacing `new_environment` with the `tmp_env` fixture instead

I realized that a little too late. Using the `tmp_env` fixture will take a little more rewriting, but it is necessary to not have the tests break.

marcoesters:
I ended up using the `conda_cli` fixture instead of `tmp_env` because the tests monitor the output stream. Menuinst exceptions do not result in a non-zero exit code, but should result in a test failure.

kenodegard (approved):

Issue #321, opened 2025-03-31
CMD shell activation changed in `conda 25.3.0`


kenodegard:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

The CMD shell activation was modified in `conda 25.3.0` to support users using Windows AppLocker with script restrictions enabled.

Xref https://github.com/conda/conda/issues/13610
Xref https://github.com/conda/conda/pull/14607

This change results in `conda shell.cmd.exe activate *` no longer producing a dynamic `.bat` script to be executed but rather produces an ini-style `.env` file to be consumed by the `CONDA_ROOT\condabin\_conda_activate.bat`.

To support `conda 25.3.0+`, menuinst's manual calls to `conda shell.cmd.exe activate *` need updating to properly handle activation, e.g.:

https://github.com/conda/menuinst/blob/7e1aa1fc445935d25f7d22cf808b68d41fa6956c/menuinst/platforms/win.py#L271-L283

This change is likely the cause of the recent failures:
- https://github.com/conda/menuinst/actions/runs/14110630226/job/39528799895?pr=319

### Conda Info

```shell

```

### Conda Config

```shell

```

### Conda list

```shell

```

kenodegard:
To catch these issues earlier in the menuinst should include conda-canary in its test matrix and probably run the tests on some regular schedule

marcoesters:
Agreed on the canaries. This is a breaking change for `menuinst` because old versions of `conda` do not use this kind of logic. So, we will have to implement a different activator depending on the `conda` version.

kenodegard:
An alternative would be for us to revert the change in `conda` and to introduce a new activator for the new format (e.g., `conda shell.env ...`)

@jezdez thoughts?

kenodegard:
yes that would shift the problem from `conda` to `menuinst`

but that would alleviate the pressure of needing to solve the problem right now for the current miniconda release

marcoesters:
If the AppLocker problem is shifted to `menuinst`, Miniconda would still not function though since the prompts are part of the installation. I think solving this problem here, too, is the best way forward.

marcoesters:
Even though, I'm not sure if any of the shortcuts use the `activate` property, so we might be getting away with it.

marcoesters:
I've thought about this more and I think this needs some sort of way to create callable files from the `conda` activator or we have to duplicate the activation code in `menuinst`.

[...]

PR #330, opened 2025-05-05
Add activator for `env` files generated by `conda`

marcoesters:
### Description

`conda` changed its activation behavior for `cmd.exe`. Starting with version `23.5.0` (see #321), `conda` produces `.env` files that are directly consumed by `conda` without providing a separate activator interface that `menuinst` can use.

This PR ports the `conda` activator as a stop-gap until `conda` provides an activator again that can be called by programs like `menuinst`.

This functionality has the disadvantage though that the behavior will freeze the `conda` version that created the shortcut. If future `conda` versions add environment variables or scripts, the shortcuts will not use the new functionality until the shortcut is re-installed. However, until an activator is provided, this will at least make `menuinst` functional.

What is puzzling is that the tests still pass on the CI whereas locally, the tests fail with the current `main` branch. The changes in this PR make the tests pass locally again.

### Checklist - did you ...


menuinst/platforms/win.py
```diff
@@ -271,14 +271,51 @@ def _command(self) -> str:
...
+                    "shell.cmd.exe",
+                    "activate",
+                    str(self.menu.prefix),
+                ]
+                activator_run = logged_run(activator_cmd, check=True, log=False)
+                activation_file = Path(activator_run.stdout.strip())
+                filetype = activation_file.suffix
+                if filetype == ".bat":
+                    activator = (
+                        f'{self.menu.conda_exe} shell.cmd.exe activate "{self.menu.prefix}"'
```
jaimergp, line 296:
```suggestion
                        f'"{self.menu.conda_exe}" shell.cmd.exe activate "{self.menu.prefix}"'
```

Is this quoted at source? If not, we should quote it.

news/330-windows-activation-env-file
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Add activator for `env` files generated by `conda`. (#321 via #330)
```
jaimergp, line 3:
```suggestion
* Add activator for `env` files generated by `conda` 25.3+. (#321 via #330)
```

tests/data/jsons/menu-name.json
```diff
@@ -13,6 +13,7 @@
         "echo",
         "A"
       ],
+      "activate": false,
```
jaimergp, line 16: Why do we need these test changes?

marcoesters: `activate` only works when the prefix is a `conda` environment, which is not the case for these shortcuts. This causes the activator command to fail. For `.bat` files, that command fails when the shortcut is executed, which we don't do for the tests.

For `.env` files, however, the activator is called at install time, which causes the installation of the shortcut to fail. I actually like that behavior more, even though it is currently limited to Windows.

menuinst/platforms/win.py
marcoesters: It is currently not quoted and I didn't find a difference in behavior whether we quote it or not. This does bring me to a different question though: we may need a separate test branch for `conda < 25.3.x` for Windows if we want to test that branch since the latest conda version will be a moving target. Thoughts?

jaimergp: We can add a new CI job with these constraints while we still support that version, `menunist` is pretty cheap CI-wise thankfully.

tests/data/jsons/menu-name.json
jaimergp: Hm, I see. I wonder if we check for `conda` in the target prefix before we write the shortcut logic, but that will require API changes so `conda` can pass that boolean, because I'm not sure `menuinst` should be the one checking. Thoughts?

menuinst/platforms/win.py
```diff
@@ -271,14 +271,51 @@ def _command(self) -> str:
...
+                    activation_lines = []
+                    for line in activation_script:
+                        keyword, value = line.split("=", 1)
+                        if keyword == "_CONDA_SCRIPT":
+                            activation_lines.append(f"@CALL {value}")
+                        else:
+                            activation_lines.append(f'@SET "{keyword}={value}"')
+                else:
+                    raise NotImplementedError(
+                        f"Menuinst cannot parse activation scripts of type {filetype}."
```
jaimergp, line 313: Let's mention the full path here too:

```suggestion
                        f"Menuinst cannot parse activation scripts of type '{filetype}': '{activation_file}'"
```

menuinst/platforms/win.py
```diff
@@ -271,14 +271,51 @@ def _command(self) -> str:
...
+                    )
+                    activation_lines = [
+                        f'@FOR /F "usebackq tokens=*" %%i IN (`{activator}`) do set "ACTIVATOR=%%i"',  # noqa
+                        "@CALL %ACTIVATOR%",
+                    ]
+                elif filetype == ".env":
+                    activation_script = activation_file.read_text().splitlines()
+                    activation_lines = []
+                    for line in activation_script:
+                        keyword, value = line.split("=", 1)
```
jaimergp, line 306:
```suggestion
                        keyword, value = line.strip().split("=", 1)
```

just in case?

menuinst/platforms/win.py
```diff
@@ -271,14 +271,51 @@ def _command(self) -> str:
...
+                        f'{self.menu.conda_exe} shell.cmd.exe activate "{self.menu.prefix}"'
+                    )
+                    activation_lines = [
+                        f'@FOR /F "usebackq tokens=*" %%i IN (`{activator}`) do set "ACTIVATOR=%%i"',  # noqa
+                        "@CALL %ACTIVATOR%",
+                    ]
+                elif filetype == ".env":
+                    activation_script = activation_file.read_text().splitlines()
+                    activation_lines = []
+                    for line in activation_script:
```
jaimergp, line 305:
```suggestion
                    activation_lines = []
                    for line in activation_file.read_text().splitlines():
```

We don't need that variable anywhere else, do we?

jaimergp (approved):

menuinst/platforms/win.py
marcoesters: Turns out that quoting is a problem on some systems. https://github.com/conda/menuinst/pull/330/commits/6de834d12fd0b9c14f5a7d7f57e998ffdc203a64 got the tests that used `bat` files working again.

menuinst/platforms/win.py
```diff
@@ -106,11 +106,11 @@ def placeholders(self) -> Dict[str, str]:
...
-            self.base_prefix / "_conda.exe",
             Path(os.environ.get("CONDA_EXE", r"C:\oops\a_file_that_does_not_exist")),
+            Path(os.environ.get("MAMBA_EXE", r"C:\oops\a_file_that_does_not_exist")),
             self.base_prefix / "condabin" / "conda.bat",
             self.base_prefix / "bin" / "conda.bat",
-            Path(os.environ.get("MAMBA_EXE", r"C:\oops\a_file_that_does_not_exist")),
+            self.base_prefix / "_conda.exe",
             self.base_prefix / "condabin" / "micromamba.exe",
             self.base_prefix / "bin" / "micromamba.exe",
         )
```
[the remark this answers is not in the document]
marcoesters: @jaimergp, after some consideration, I don't think we should break the existing behavior, so I reverted that change. Staying with a stable `conda` version (i.e., `conda-standalone`) is a good idea and probably prevented this problem flying in our face already.

That these changes do what they need to do has been shown here: https://github.com/conda/menuinst/pull/330/commits/9556e7ef8c96f97539afbffa79a302f7d9ca207a

Miniforge will eventually catch up and release an installer with a newer `conda-standalone` version.

jaimergp (approved):

PR #331, opened 2025-05-06
Log at debug level when an exception occurs for winreg.Openkey

mrclary:
### Description

Log at debug level when an exception occurs for winreg.Openkey. See spyder-ide/spyder#24339

### Checklist - did you ...

- [ ] ~Add a file to the `news` directory ([using the template](https://github.com/conda/menuinst/blob/main/news/TEMPLATE)) for the next release's release notes?~
     
- [ ] ~Add / update necessary tests?~
- [ ] ~Add / update outdated documentation?~

### DIFF
--- a/menuinst/platforms/win_utils/registry.py
+++ b/menuinst/platforms/win_utils/registry.py
@@ -109,9 +109,7 @@ def unregister_file_extension(extension, identifier, mode="user"):
         ) as key:
             winreg.DeleteValue(key, identifier)
     except FileNotFoundError:
-        log.debug(
-            "Handler '%s' is not associated with extension '%s'", identifier, extension
-        )
+        log.debug("Handler '%s' is not associated with extension '%s'", identifier, extension)
     except Exception as exc:
         log.debug("Could not check key '%s' for deletion", extension, exc_info=exc)
         return False
--- a/menuinst/platforms/win_utils/registry.py
+++ b/menuinst/platforms/win_utils/registry.py
@@ -116,7 +116,7 @@ def unregister_file_extension(extension, identifier, mode="user"):
             else:
                 winreg.DeleteValue(key, identifier)
     except Exception as exc:
-        log.exception("Could not check key '%s' for deletion", extension, exc_info=exc)
+        log.debug("Could not check key '%s' for deletion", extension)
         return False
 
 
--- a/menuinst/platforms/win_utils/registry.py
+++ b/menuinst/platforms/win_utils/registry.py
@@ -107,10 +107,13 @@ def unregister_file_extension(extension, identifier, mode="user"):
         with winreg.OpenKey(
             root, rf"Software\Classes\{extension}\OpenWithProgids", 0, winreg.KEY_ALL_ACCESS
         ) as key:
-            winreg.QueryValueEx(key, identifier)  # ??? Is this necessary?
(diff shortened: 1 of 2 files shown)


travishathaway:
@conda-bot check

marcoesters:
I agree that the exception output is not appropriate in the issue your user reported. The error (`FileNotFoundError`) indicates that the registry key doesn't exist in the first place.

It doesn't necessarily have to be a permission error. I'm wondering if the initial installation was done without calling `menuinst` - a subsequent `conda install` without `--no-shortcuts` may try and unregister the handler, which doesn't exist. I'm not 100% sure if that's what happens though.

So, my suggestion is to remove the [inner try-except](https://github.com/conda/menuinst/blob/1866363f197e0633fae0db8569119d102ca8d9cc/menuinst/platforms/win_utils/registry.py#L110-L117) block and keep the debug log output on `FileNotFoundError` and keep the exception output for all other errors.

As an aside, it looks like the function also has an inconsistent return type: it's either `False` or `None`, but can never be `True`.

marcoesters:
pre-commit.ci autofix

menuinst/platforms/win_utils/registry.py
```diff
@@ -107,16 +107,10 @@ def unregister_file_extension(extension, identifier, mode="user"):
...
-                log.debug(
-                    "Handler '%s' is not associated with extension '%s'", identifier, extension
-                )
-            else:
-                winreg.DeleteValue(key, identifier)
-    except Exception as exc:
-        log.exception("Could not check key '%s' for deletion", extension, exc_info=exc)
+            winreg.QueryValueEx(key, identifier)  #??? Is this necessary?
+            winreg.DeleteValue(key, identifier)
+    except Exception:
```
marcoesters, line 112: I think the `FileNotFoundError` exeption should be added here. All other errors can still look like a traceback.

menuinst/platforms/win_utils/registry.py
```diff
@@ -107,16 +107,10 @@ def unregister_file_extension(extension, identifier, mode="user"):
...
-                winreg.QueryValueEx(key, identifier)
-            except FileNotFoundError:
-                log.debug(
-                    "Handler '%s' is not associated with extension '%s'", identifier, extension
-                )
-            else:
-                winreg.DeleteValue(key, identifier)
-    except Exception as exc:
-        log.exception("Could not check key '%s' for deletion", extension, exc_info=exc)
+            winreg.QueryValueEx(key, identifier)  # ??? Is this necessary?
```
marcoesters, line 110: This is probably not necessary since `winreg.DeleteValue` already raises an `OSError`.

Could you add a test for this inside `test_windows_registry.py`? The `capsys` fixture should be good to capture the output.

mrclary: You mean to test for `OSError`? Not sure what "test for this" refers to.

marcoesters: That's right. It looks like [tests/test_windows_registry.py](https://github.com/conda/menuinst/blob/d0fce331068023722aece0c5d6a1479b20c5ea29/tests/test_windows_registry.py) has tests for successful runs, but not for failures.

marcoesters (approved):
Thank you!

PR #340, opened 2025-06-04
use XDG_MENU_PREFIX where appropriate

### LINKED ISSUE #339: Is there a way to put Linux menu shortcuts into "Science" or "Development" categories?
### Checklist

### What is the idea?

Over at https://github.com/scientific-python/installer/issues/11 we noticed that Linux desktop application shortcuts go into the "Other" category by default.  Is there a way that we missed to specify another category, such as "Science" or "Development" for our application shortcuts on Linux?  If not, do we need to move the files in a post-install step, or is there a better way?

### Why is this needed?

_No response_

### What should happen?

_No response_

### Additional Context

_No response_

drammock:
### Description

Fixes #339.

xref https://specifications.freedesktop.org/menu-spec/1.0/paths.html

### Checklist - did you ...

- [ ] ~~Add / update outdated documentation?~~ *there isn't any documentation discussing the `applications.menu` file*

### DIFF
--- a/tests/test_conda.py
+++ b/tests/test_conda.py
@@ -83,7 +83,12 @@ def test_conda_recent_enough():
 
 @pytest.mark.skipif(PLATFORM != "linux", reason="Linux only")
 def test_package_1_linux(tmpdir, conda_cli):
-    applications_menu = Path(tmpdir) / "config" / "menus" / "applications.menu"
+    applications_menu = (
+        Path(tmpdir)
+        / "config"
+        / "menus"
+        / f"{os.environ.get('XDG_MENU_PREFIX', '')}applications.menu"
+    )
     if applications_menu.is_file():
         original_xml = applications_menu.read_text()
     else:
--- a/menuinst/platforms/linux.py
+++ b/menuinst/platforms/linux.py
@@ -43,10 +43,9 @@ class LinuxMenu(Menu):
             ).expanduser()
 
         # XML Config paths
-        self.system_menu_config_location = (
-            self._system_config_directory / "menus" / "applications.menu"
-        )
-        self.menu_config_location = self.config_directory / "menus" / "applications.menu"
+        menu_fname = f"{os.environ.get('XDG_MENU_PREFIX', '')}applications.menu"
+        self.system_menu_config_location = self._system_config_directory / "menus" / menu_fname
+        self.menu_config_location = self.config_directory / "menus" / menu_fname
         # .desktop / .directory paths
         self.directory_entry_location = (
             self.data_directory
--- a/news/340-use-xdg-menu-prefix
+++ b/news/340-use-xdg-menu-prefix
@@ -1,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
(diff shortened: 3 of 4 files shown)


jaimergp (approved):
LGTM, I'll merge once the CLA checks are passing. Thanks!

tests/test_conda.py
```diff
@@ -83,7 +83,12 @@ def test_conda_recent_enough():
 
 @pytest.mark.skipif(PLATFORM != "linux", reason="Linux only")
 def test_package_1_linux(tmpdir, conda_cli):
-    applications_menu = Path(tmpdir) / "config" / "menus" / "applications.menu"
+    applications_menu = (
+        Path(tmpdir)
+        / "config"
+        / "menus"
+        / f"{os.environ.get('XDG_MENU_PREFIX', '')}applications.menu"
```
marcoesters, line 90: I think instead of this, I'd monkeypatch `XDG_MENU_PREFIX` with a known value.

drammock: you mean instead of providing the fallback `''` value?  What would be the benefit of doing it that way (unless you *also* add a check that the resulting filename correctly contains `XDG_MENU_PREFIX`)?

I didn't look too closely at this test before, beyond searching the repo for `applications.menu` and making sure the tests still passed. I tried digging a bit deeper but something doesn't add up for me: the test is checking for the presence of `applications.menu` before `install_package_1` is getting called, and it's looking *in a tempdir* (not in the usual system places) so we wouldn't expect there to be existing `applications.menu` files there anyway.  Thus in my local tests, `applications.menu` *never* exists at the point when it's looked for (and thus `original_xml` is always `None`). So I don't understand why the if/else branching around `applications_menu.isfile()` is even there.

As an aside, when I run with `pytest -s` I see output like

    ```
DEBUG:menuinst.platforms.linux:Creating /tmp/pytest-of-drmccloy/pytest-12/test_package_1_linux0/config/menus/xfce-applications.menu
DEBUG:menuinst.platforms.linux:Editing /tmp/pytest-of-drmccloy/pytest-12/test_package_1_linux0/config/menus/xfce-applications.menu to add Package 1 config
    ```

    ...suggesting that `menuinst` is doing the right thing (creating the file with the correct name). But AFAICT because this test checks for the existence of the file *before* the package is installed, it won't actually test whether `menuinst` is in fact naming the file correctly.

marcoesters: My apologies, I made the comment in the wrong file. This would have been a more appropriate place for the comment where the file at least gets parsed and you can test that the environment variable gets correctly propagated: https://github.com/conda/menuinst/blob/5b3fbf38bd5b23ab7d2ecaea126122849ecbf667/tests/test_api.py#L217-L221

I won't block merging this PR for a missing monkeypatch though.

(side note: @jaimergp, we should probably generally patch the `XDG_*` values so that the tests do not write into an existing user/system configuration)

>  Thus in my local tests, applications.menu never exists at the point when it's looked for (and thus original_xml is always None). So I don't understand why the if/else branching around applications_menu.isfile() is even there.

I think I agree with you here. While you can set the prefix of the temporary pytest directory, I can't say I can come up with a way where you could point `tmpdir` to the actual `XDG_*` directories.

Issue #296, opened 2025-01-16
on a pure ubuntu server: FileNotFoundError: [Errno 2] No such file or directory: 'xdg-mim


ReimarBauer:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

When I install mss on a fresh ubuntu server I get a 

`mamba create -n mssenv mss==9.3.0`
```
Preparing transaction: done
Verifying transaction: done
Executing transaction: | menuinst Exception
Traceback (most recent call last):
  File "/home/user/miniforge3/lib/python3.12/site-packages/conda/gateways/disk/create.py", line 261, in make_menu
    menuinst.install(
  File "/home/user/miniforge3/lib/python3.12/site-packages/menuinst/api.py", line 170, in _install_adapter
    install(metadata, target_prefix=prefix, **kwargs)
  File "/home/user/miniforge3/lib/python3.12/site-packages/menuinst/utils.py", line 426, in wrapper_elevate
    return func(
           ^^^^^
  File "/home/user/miniforge3/lib/python3.12/site-packages/menuinst/api.py", line 62, in install
    paths += menu_item.create()
             ^^^^^^^^^^^^^^^^^^
  File "/home/user/miniforge3/lib/python3.12/site-packages/menuinst/platforms/linux.py", line 198, in create
    self._maybe_register_mime_types(register=True)
  File "/home/user/miniforge3/lib/python3.12/site-packages/menuinst/platforms/linux.py", line 280, in _maybe_register_mime_types
    self._register_mime_types(mime_types, register=register)
  File "/home/user/miniforge3/lib/python3.12/site-packages/menuinst/platforms/linux.py", line 287, in _register_mime_types
    self._glob_pattern_for_mime_type(mime_type, glob_pattern, install=register)
  File "/home/user/miniforge3/lib/python3.12/site-packages/menuinst/platforms/linux.py", line 380, in _glob_pattern_for_mime_type
    logged_run(
  File "/home/user/miniforge3/lib/python3.12/site-packages/menuinst/utils.py", line 464, in logged_run
    process = subprocess.run(args, capture_output=True, text=True, **kwargs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/miniforge3/lib/python3.12/subprocess.py", line 548, in run
    with Popen(*popenargs, **kwargs) as process:
[...]

PR #403, opened 2025-10-29
Improve handling of MimeType with glob patterns when xdg-mime is not present

lrandersson:
### Description

- To resolve https://github.com/conda/menuinst/issues/296
- Ideally the presence of `xdg-mime` could be checked elsewhere, however there are more and similar uses of `shutil.which` in that same file and therefore I went with a similar approach. A refactoring could be done to optimize the code (to minimize the amount of checks via `shutil.which`) but I think the gain is negligble.
- Note that there was already a fallback implemented, but this only triggered if there was a `CalledProcessError` and would not trigger if the program itself (xdg-mime) did not exist.

### Checklist - did you ...

### DIFF
--- a/menuinst/platforms/linux.py
+++ b/menuinst/platforms/linux.py
@@ -389,9 +389,11 @@ class LinuxMenuItem(MenuItem):
                 else:
                     name = self.render_key("name")
                     log.warning(
-                        f"Unable to un/register MIME type '{glob_pattern}' "
-                        f"for Desktop Entry of name '{name}', "
-                        "'xdg-mime' is not present on the system.'"
+                        "Unable to un/register MIME type '%s' "
+                        "for Desktop Entry of name '%s': "
+                        "'xdg-mime' is not present on the system.'",
+                        glob_pattern,
+                        name,
                     )
                     use_fallback = True
         except CalledProcessError:
--- a/news/403-improve-error-handling-xdg-mime
+++ b/news/403-improve-error-handling-xdg-mime
@@ -4,7 +4,7 @@
 
 ### Bug fixes
 
-* For Linux platforms, a MimeType with defined glob patterns are now properly handled when 'xdg-mime' is not on the system. (#290 via #403)
+* For Linux platforms, MimeTypes with defined glob patterns are now properly handled when `xdg-mime` is not on the system. (#290 via #403)
 
 ### Deprecations
 


news/403-improve-error-handling-xdg-mime
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* For Linux platforms, a MimeType with defined glob patterns are now properly handled when 'xdg-mime' is not on the system. (#290 via #403)
```
marcoesters:
```suggestion
* For Linux platforms, a MimeType with defined glob patterns are now properly handled when `xdg-mime` is not on the system. (#290 via #403)
```

menuinst/platforms/linux.py
```diff
@@ -374,20 +374,34 @@ def _glob_pattern_for_mime_type(
...
+                        [xdg_mime, subcommand, "--mode", self.menu.mode, "--novendor", f.name],
+                        check=True,
+                    )
+                else:
+                    name = self.render_key("name")
+                    log.warning(
+                        f"Unable to un/register MIME type '{glob_pattern}' "
+                        f"for Desktop Entry of name '{name}', "
+                        "'xdg-mime' is not present on the system.'"
+                    )
```
marcoesters: f-strings in logging are generally discouraged. Python's `logging` module has built-in formatting functionality that is only called when needed: https://docs.astral.sh/ruff/rules/logging-f-string/

menuinst/platforms/linux.py
```diff
@@ -374,20 +374,34 @@ def _glob_pattern_for_mime_type(
...
+                    logged_run(
+                        [xdg_mime, subcommand, "--mode", self.menu.mode, "--novendor", f.name],
+                        check=True,
+                    )
+                else:
+                    name = self.render_key("name")
+                    log.warning(
+                        f"Unable to un/register MIME type '{glob_pattern}' "
+                        f"for Desktop Entry of name '{name}', "
+                        "'xdg-mime' is not present on the system.'"
```
marcoesters:
```suggestion
                        f"Unable to un/register MIME type '{glob_pattern}' "
                        f"for Desktop Entry of name '{name}': "
                        "'xdg-mime' is not present on the system.'"
```

marcoesters (commented):
Looks good, just minor comments.

news/403-improve-error-handling-xdg-mime
lrandersson: Fixed

PR #415, opened 2025-11-25
Allow arbitrary properties in `Info.plist` files

### LINKED ISSUE #195: Expand Info Property List keywords for MacOS apps
### Checklist

### What is the idea?

The keyword list for the Information Property List (`Info.plist`) files is very limited.

This can be done in one of two ways:

* Add all keywords to the schema. This allows for type validation, but is not scalable as the list needs to be kept up-to-date with Apple. Additionally, many keywords may not be relevant for `menuinst`, so adding all keywords may not be worth the effort.
* Add a free-form item to the schema (e.g., `info_plist`). This is scales well and invalid keywords do not break the app, according to my tests. However, type validation is not possible with this approach. This could be solved if a standard was published somewhere, but I'm not aware of any such document.

### Why is this needed?

The current list of keywords for `Info.plist` is limited and new keywords require an updated schema (and possibly CEP) to be published. See, e.g., https://github.com/conda/menuinst/pull/194 for how adding new keywords currently needs to be performed.

### What should happen?

The schema should be expanded to allow for any valid `Info.plist` keyword for MacOS apps.

### Additional Context

_No response_

### LINKED ISSUE #410: Help resolving issues in a MacOS-Shortcut bundled using `menuinst`
I help maintain a [feedstock](https://github.com/conda-forge/alacritty-feedstock) for a Terminal Emulator called Alacritty, recently I received an [issue](https://github.com/conda-forge/alacritty-feedstock/issues/6) regarding the app bundled for MacOS, the issues are the following,
- Alacritty not inheriting the System-default for the Theme
- Alacritty not maximizing to cover the whole page
- The menu option _Alacritty -> About Alacritty_ not showing the version number

_(the issue page has videos for all of the issues)_

---

Doing the following resolves all of the issues mentioned above,
- Replacing the contents of `~/Applications/Alacritty.app/Contents/MacOS/` with the actual binary that gets called by the shell script (`alacritty-script`)
- Adding the following entries to the `Info.plist` file
```xml
  <key>NSRequiresAquaSystemAppearance</key>
  <string>NO</string>
  <key>NSHighResolutionCapable</key>
  <true/>
```

---

I was hoping I could get some help resolving these issues directly via `menuinst` and not rely on manually manipulating the installed files.

This is the file I used to configure `menuinst` for Alacritty: [menu.json](https://github.com/conda-forge/alacritty-feedstock/blob/main/recipe/menu.json), I have been going through the [menuinst-reference](https://conda.github.io/menuinst/reference/) to understand if I have missed something, but haven't made any breakthroughs.

I was hoping you would be able to offer help, that might steer me in the right direction. I would greatly appreciate any and all assistance.

marcoesters:
### Description

`Info.plist` files contain properties for macOS app bundles. `menuinst` currently supports only a handful of these. Each new properties require a schema update, which does not scale. Moreover, app bundles that contain properties that macOS does not recognize are not invalid and function normally.

Instead of continuously adding new properties, add a property that allows shortcut authors to define any `Info.plist` property. In the future, `menuinst` should only explicitly add those that it needs to manipulate and leave other properties to the extra keyword list. Old keywords remain for now to maintain backwards compatibility.

This updates the schema version.

Closes #195 
Closes #410 

### Checklist - did you ...

### DIFF
--- a/menuinst/platforms/osx.py
+++ b/menuinst/platforms/osx.py
@@ -164,7 +164,7 @@ class MacOSMenuItem(MenuItem):
 
         # Override defaults with (potentially) user provided values
         ignore_keys = (*menuitem_defaults, "entitlements", "link_in_bundle", "info_plist_extra")
-        info_plist_extra = self.metadata.get("info_plist_extra", {})
+        info_plist_extra = self.metadata.get("info_plist_extra") or {}
         for key in menuitem_defaults["platforms"]["osx"]:
             if key in ignore_keys:
                 continue


XhstormR:
ping

marcoesters:
> > In the future, menuinst should only explicitly add those that it needs to manipulate and leave other properties to the extra keyword list. Old keywords remain for now to maintain backwards compatibility.
> 
> I'd like to see this taken care of in another PR, where we deprecate the top-level plist keys we don't need in favor of `extra_plist_info`.

I created #419 for this.

PR #454, opened 2026-03-25
Decorator elevate_as_needed now respects .nonadmin

### LINKED ISSUE #453: [Windows] @elevate_as_needed decorator ignores .nonadmin marker when user is admin
### Checklist

### What happened?

On Windows, when running as admin with a `.nonadmin` marker file present within the install directory, the new-style menu JSON files (those with `$schema`) create shortcuts in the system location (ProgramData) instead of the user location (AppData).

**Expected behavior:** The `.nonadmin` marker file should signal a user-mode install regardless of whether the process is elevated.

Current behavior:
- Legacy code path (no `$schema`): Checks `.nonadmin` before deciding mode (which seems to work as expected)
- New code path (with `$schema`): `@elevate_as_needed` decorator checks `user_is_admin()` first, passes _mode="system" without checking `.nonadmin` (which does not work as expected)

Reproduction (`conda install spyder` shows the bug):
1. Create a conda environment with `.nonadmin` marker file within `PREFIX`
2. Run conda install as an elevated process (e.g., "Run as Administrator")
3. Install a package with new-style menu JSON (`$schema` present)
4. Observe: shortcut created in `%ProgramData%` instead of `%AppData%`

Suspected location: `menuinst/utils.py` lines 509-517, the `@elevate_as_needed` decorator:
```
if user_is_admin():
   return func(*args, _mode="system", **kwargs)
```
This should check for `.nonadmin` before defaulting to system mode.

### Conda Info

```shell

```

### Conda Config

```shell

```

### Conda list

```shell

```

### Additional Context

The legacy menuinst code (`_legacy/__init__.py` lines 62-87) correctly checks `.nonadmin` before deciding the mode. The new code path should have consistent behavior.

This inconsistency causes issues when `conda` is invoked from an elevated process but the install is intended to be per-user (as signaled by `.nonadmin`).

lrandersson:
### Description

This PR resolves: https://github.com/conda/menuinst/issues/453.
The `@elevate_as_needed` decorator ignored the `.nonadmin` marker file when `user_is_admin()` returned `True`, always using system mode.
This caused shortcuts to be created in the system location (ProgramData) instead of the user location (AppData) for per-user installs running in an elevated context.

The fix checks for `.nonadmin` before defaulting to system mode when admin.

### Checklist - did you ...

### DIFF
--- a/menuinst/utils.py
+++ b/menuinst/utils.py
@@ -507,11 +507,16 @@ def elevate_as_needed(func: Callable) -> Callable:
                 if return_code == 0:  # success, we are done
                     return
         elif user_is_admin():
-            # We are already running as admin, no need to elevate
+            # We are already running as admin, but check if .nonadmin marker exists
+            # which signals a user-mode install (e.g., per-user MSI install running elevated)
+            nonadmin_exists = (
+                Path(target_prefix, ".nonadmin").is_file()
+                or Path(base_prefix, ".nonadmin").is_file()
+            )
             return func(
                 target_prefix=target_prefix,
                 base_prefix=base_prefix,
-                _mode="system",
+                _mode="user" if nonadmin_exists else "system",
                 *args,
                 **kwargs,
             )
--- a/menuinst/utils.py
+++ b/menuinst/utils.py
@@ -508,7 +508,7 @@ def elevate_as_needed(func: Callable) -> Callable:
                     return
         elif user_is_admin():
             # We are already running as admin, but check if .nonadmin marker exists
-            # which signals a user-mode install (e.g., per-user MSI install running elevated)
+            # which signals a user-mode install.
             nonadmin_exists = (
                 Path(target_prefix, ".nonadmin").is_file()
                 or Path(base_prefix, ".nonadmin").is_file()
--- a/menuinst/utils.py
+++ b/menuinst/utils.py
@@ -507,8 +507,7 @@ def elevate_as_needed(func: Callable) -> Callable:
                 if return_code == 0:  # success, we are done
                     return
         elif user_is_admin():
-            # We are already running as admin, but check if .nonadmin marker exists
-            # which signals a user-mode install.
+            # On Windows, check if .nonadmin marker exists which signals a user-mode install.
             nonadmin_exists = (
                 Path(target_prefix, ".nonadmin").is_file()
                 or Path(base_prefix, ".nonadmin").is_file()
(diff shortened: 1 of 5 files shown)


jaimergp (dismissed):

menuinst/utils.py
```diff
@@ -507,11 +507,16 @@ def wrapper_elevate(
                 if return_code == 0:  # success, we are done
                     return
         elif user_is_admin():
-            # We are already running as admin, no need to elevate
+            # We are already running as admin, but check if .nonadmin marker exists
+            # which signals a user-mode install.
+            nonadmin_exists = (
```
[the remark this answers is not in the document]
marcoesters: I think to test this, we need to check what happens when `Administrator`/`root` installs a shortcut with a `.nonadmin` file in prefix, then try to uninstall it as a single user later.

lrandersson:
@jaimergp after I had a discussion with @marcoesters we changed the implementation to be Windows only. 
For both of you reviewers, I also added this commit because I found some places where we did not quote on Windows potential paths with spaces. https://github.com/conda/menuinst/pull/454/changes/3efa79f7568c996f3b1154139ca926b314dbb9f4

On Linux, if you install with sudo/root privileges, you're expected to also remove the package with sudo/root. Otherwise you will end-up with permission related issues. For example, with this fix, if you run `conda install ...` as a `root` user, the shortcuts will still be installed into a directory inaccessible to non-admin users. Therefore with/without this change, you will have permission related issues and we decided to go with the more conservative approach and simply update it only on Windows.

marcoesters (approved):
I think that works and makes shortcuts on Windows more predictable. I'd like to get @jaimergp's input though before merging.

lrandersson:
> I think that works and makes shortcuts on Windows more predictable. I'd like to get @jaimergp's input though before merging.

ping @jaimergp

tests/test_windows_menu_item.py
```diff
@@ -117,6 +117,87 @@ def test_process_command_as_activated_and_terminal(with_arg1):
...
+    )
+    fake_env_file.write_text(fake_env_content)
+
+    # Create a menu item with activate=True to trigger the activation code path
+    class TestMenu(WindowsMenu):
+        def __init__(self):
+            self.mode = "user"
+            self.name = "Test"
+            self.prefix = Path(prefix_path)
+            self.base_prefix = Path(prefix_path).parent
```
jaimergp, line 151: This is usually two parents up (because of the `envs/` directory). This doesn't matter for the test, though, right?

jaimergp (approved):
LGTM!

tests/test_windows_menu_item.py
lrandersson: Good point, and yeah it does not matter for the purpose of testing since we are more or less just "mocking" it.

PR #475, opened 2026-04-30
Fix conditional platforms and logging

### LINKED ISSUE #463: KeyError in Windows install for shortcut only defined for Linux
On latest stable `menuinst=2.4.2` and `conda=26.1.1` (26.3.1 seems to have a separate JSON export issue on Windows 😓 ), [this CI run installing a `constructor`-built `.exe`](https://github.com/mne-tools/mne-installers/actions/runs/24151954276/job/70483189204#step:5:40) successfully installs our locally-built `mne-installer-menus` package:

<details>
<summary>Successful install log</summary>

```
MNE-Python will now be installed into this location:
C:\Users\runneradmin\mne-python\1.12.0_0
Unpacking payload...
Setting up the package cache...
Setting up the base environment...
Installing packages for base, creating shortcuts if necessary...
Preparing transaction:
...working...
done
Executing transaction:
...working...




menuinst\platforms\win.py:72: UserWarning: Quick launch menus are not available for system level installs
Terminal profiles are not available for system level installs
Terminal profiles are not available for system level installs
Terminal profiles are not available for system level installs
Terminal profiles are not available for system level installs
done
```

</details>

So far, so good. However, when I [add](https://github.com/mne-tools/mne-installers/pull/430/changes/25bbff67a9efed00f0b8d64ddaa21d999933342f#diff-9416bf1f7622b2b406df4eadae3d565d2be5eed27f9796e186c1ec9a5caadd18R145) an additional Linux-only entry to the `mne-installer-menu` package that gets installed:

<details>
<summary>Additional shortcut</summary>

```
    },
    {
      "name": "MNE-Python",
      "description": "MNE-Python generic launcher",
      "icon": "{{ MENU_DIR }}/mne_default_icon.png",
      "activate": true,
      "terminal": false,
      "command": [
        "will be overridden in platforms section"
      ],
      "platforms": {
        "linux": {
          "command": [
            "exec",
            "bash",
            "--init-file",
            "{{ MENU_DIR }}/mne_open_prompt.sh"
          ],
          "Categories": [
            "Science"
          ],
          "SingleMainWindow": true,
          "NoDisplay": true,
          "StartupWMClass": "MNE-Python"
        }
      }
```

</details>

In the [CI run](https://github.com/mne-tools/mne-installers/actions/runs/24162364418/job/70517871604?pr=430#step:5:37) there is a (non-fatal) error logged by `menuinst` because it hits a `KeyError: 'desktop'`:

<details>
<summary>Errant install log</summary>

```
MNE-Python will now be installed into this location:
C:\Users\runneradmin\mne-python\1.12.0_0
Unpacking payload...
Setting up the package cache...
Setting up the base environment...
Installing packages for base, creating shortcuts if necessary...
Preparing transaction:
...working...
done
Executing transaction:
...working...
menuinst\platforms\win.py:72: UserWarning: Quick launch menus are not available for system level installs
Terminal profiles are not available for system level installs
Terminal profiles are not available for system level installs
Terminal profiles are not available for system level installs
Terminal profiles are not available for system level installs
menuinst Exception
  File "conda\gateways\disk\create.py", line 241, in make_menu
  File "menuinst\api.py", line 191, in _install_adapter
  File "menuinst\utils.py", line 423, in wrapper_elevate
  File "menuinst\api.py", line 77, in install
  File "menuinst\platforms\win.py", line 162, in create
  File "menuinst\platforms\win.py", line 228, in _paths
KeyError: 'desktop'
done
```

</details>

(It seems like it shouldn't try to install this shortcut because there is no `"windows"` in the `platforms` section, so hitting an error here doesn't make sense to me...)

As an additional side note, it would be really great if some of the logging messages -- including the `menuinst Exception` and `Quick launch menus are not available...` -- gave the package + `'name'` of the shortcut(s) they are referring to. Would have made this a bit easier to try to track down! In my case I was lucky because all other shortcuts *are* for Windows and all contained a `desktop` entry, so it was easy to identify the problematic one...

### LINKED ISSUE #464: Add shortcut file name to logging and exception outputs
### Checklist

### What is the idea?

Logging messages and exception tracebacks should provide information on which shortcut they are referring to.

### Why is this needed?

Users often install multiple packages with shortcuts. Adding shortcut identifiers will make debugging much easier in that situation.

### What should happen?

Tracebacks and logging outputs should provide information on which shortcut is being processed.

### Additional Context

_No response_

larsoner:
### Description

Closes #463
Closes #464

Turns out the issue was a gap in testing and logic: current tests were only "all plugins enabled for the same set of platforms" (which could be all platforms, or a subset of platforms). Once I added a test for "some plugins enabled for not all platforms, some enabled for all platforms" the problem became clear.

I also added logging of the menu and menuitem name in the warnings.

### Checklist - did you ...

- [x] ~~Add / update outdated documentation?~~

Related PR also opened for `conda` in https://github.com/conda/conda/pull/16000

### DIFF
--- a/menuinst/api.py
+++ b/menuinst/api.py
@@ -74,7 +74,8 @@ def install(
     paths = []
     paths += menu.create()
     for menu_item in menu_items:
-        paths += menu_item.create()
+        if menu_item.enabled_for_platform():
+            paths += menu_item.create()
 
     return paths
 
@@ -96,7 +97,8 @@ def remove(
 
     paths = []
     for menu_item in menu_items:
-        paths += menu_item.remove()
+        if menu_item.enabled_for_platform():
+            paths += menu_item.remove()
     paths += menu.remove()
 
     if not paths and _maybe_try_user(target_prefix, base_prefix):
--- a/menuinst/api.py
+++ b/menuinst/api.py
@@ -67,15 +67,15 @@ def install(
     target_prefix = target_prefix or DEFAULT_PREFIX
     base_prefix = base_prefix or DEFAULT_BASE_PREFIX
     menu, menu_items = _load(metadata_or_path, target_prefix, base_prefix, _mode)
-    if not any(item.enabled_for_platform() for item in menu_items):
+    menu_items = [item for item in menu_items if item.enabled_for_platform()]
+    if not menu_items:
         warnings.warn(f"Metadata for {menu.name} is not enabled for {sys.platform}")
         return []
 
     paths = []
     paths += menu.create()
     for menu_item in menu_items:
-        if menu_item.enabled_for_platform():
-            paths += menu_item.create()
+        paths += menu_item.create()
 
     return paths
 
@@ -91,21 +91,21 @@ def remove(
     target_prefix = target_prefix or DEFAULT_PREFIX
     base_prefix = base_prefix or DEFAULT_BASE_PREFIX
     menu, menu_items = _load(metadata_or_path, target_prefix, base_prefix, _mode)
-    if not any(item.enabled_for_platform() for item in menu_items):
+    menu_items = [item for item in menu_items if item.enabled_for_platform()]
+    if not menu_items:
         warnings.warn(f"Metadata for {menu.name} is not enabled for {sys.platform}")
         return []
 
     paths = []
     for menu_item in menu_items:
-        if menu_item.enabled_for_platform():
-            paths += menu_item.remove()
+        paths += menu_item.remove()
     paths += menu.remove()
 
     if not paths and _maybe_try_user(target_prefix, base_prefix):
         menu, menu_items = _load(metadata_or_path, target_prefix, base_prefix, "user")
+        menu_items = [item for item in menu_items if item.enabled_for_platform()]
         for menu_item in menu_items:
-            if menu_item.enabled_for_platform():
-                paths += menu_item.remove()
+            paths += menu_item.remove()
         paths += menu.remove()
 
     return paths
--- a/menuinst/api.py
+++ b/menuinst/api.py
@@ -104,7 +104,8 @@ def remove(
     if not paths and _maybe_try_user(target_prefix, base_prefix):
         menu, menu_items = _load(metadata_or_path, target_prefix, base_prefix, "user")
         for menu_item in menu_items:
-            paths += menu_item.remove()
+            if menu_item.enabled_for_platform():
+                paths += menu_item.remove()
         paths += menu.remove()
 
     return paths
--- a/menuinst/platforms/win.py
+++ b/menuinst/platforms/win.py
@@ -86,7 +86,7 @@ class WindowsMenu(Menu):
         not when it is installed.
         """
         if self.mode == "system":
-            log.warning("Terminal profiles are not available for system level installs")
+            log.warning("%s: Terminal profiles are not available for system level installs", self.name)
             return []
         return windows_terminal_settings_files(self.mode)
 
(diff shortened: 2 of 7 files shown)


tests/test_api.py
```diff
@@ -526,6 +527,38 @@ def test_platforms():
     assert not menu_item.enabled_for_platform("linux")
     assert not menu_item.enabled_for_platform("darwin")
 
+    # And now some more complete tests:
+
+    # No platforms
+    datafile = str(DATA / "jsons" / "no-platforms.json")
```
larsoner, line 532: A `git grep no-platforms` suggests that this existing file was unused, so I added a little test for it.

menuinst/platforms/win.py
```diff
@@ -422,7 +424,7 @@ def _add_remove_windows_terminal_profile(self, location: Path, remove: bool = Fa
                     settings["profiles"]["list"] = []
                 settings["profiles"]["list"].append(profile_data)
             else:
-                log.warning(f"Overwriting terminal profile for {name}.")
+                log.warning(f"{self._log_name}: Overwriting terminal profile for {name}.")
```
lrandersson:
```suggestion
                log.warning("%s: Overwriting terminal profile for %s.", self._log_name, name)
```
I see that it was an f-string before your change but just for consistency with above.

menuinst/api.py
```diff
@@ -96,7 +97,8 @@ def remove(
 
     paths = []
     for menu_item in menu_items:
-        paths += menu_item.remove()
+        if menu_item.enabled_for_platform():
+            paths += menu_item.remove()
```
lrandersson: I think we need to the same logic also in the code below this part where we currently have:
```
    if not paths and _maybe_try_user(target_prefix, base_prefix):
        menu, menu_items = _load(metadata_or_path, target_prefix, base_prefix, "user")
        for menu_item in menu_items:
            paths += menu_item.remove()
        paths += menu.remove()
```
It might not be that easy to test either because I think the tests are not running as an admin and `user_is_admin` needs to be mocked.

lrandersson (commented):
Thanks for your fix @larsoner (Go blue!), I just have some minor comments and I think in general this looks good!

menuinst/api.py
larsoner: Ahh I see, I'll at least push the fix

news/475-platform-conditional
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Improved logging of warnings during installation (#475)
```
marcoesters:
```suggestion
* Improved logging of warnings during installation (#464 via #475)
```

news/475-platform-conditional
```diff
@@ -0,0 +1,19 @@
...
+
+### Bug fixes
+
+* Fix bug where menus containing some platform-specific shortcuts would lead to install and removal errors (#475)
```
marcoesters:
```suggestion
* Fix bug where menus containing some platform-specific shortcuts would lead to install and removal errors (#463 via #475)
```

menuinst/api.py
marcoesters: Alternatively, instead of adding the check everywhere, we could filter out the menu items at the beginning of the function:

```
menu, menu_items = _load(metadata_or_path, target_prefix, base_prefix, _mode)
menu_items = [item.enabled_for_platform() for item in menu_items]
if not menu_items:
        warnings.warn(f"Metadata for {menu.name} is not enabled for {sys.platform}")
        return []
```

tests/test_api.py
marcoesters: I like the tests, but can we write separate functions for each scenario? That would make individual failures easier to spot.

marcoesters (commented):
Very nice, thank you for the fix! Just a few minor comments.

menuinst/api.py
larsoner: Yeah that's much cleaner!

larsoner:
Pushed the suggested changes, thanks for the quick reviews @lrandersson @marcoesters !

tests/test_api.py
```diff
@@ -526,6 +527,40 @@ def test_platforms():
     assert not menu_item.enabled_for_platform("linux")
     assert not menu_item.enabled_for_platform("darwin")
 
+    # And now some more complete tests:
+
```
marcoesters:
```suggestion
```

marcoesters (approved):
LGTM, assuming the tests pass.

lrandersson (approved):

PR #477, opened 2026-05-11
Add support for menuinst.toml

lrandersson:
## Summary

Adds the possibility to override `DISTRIBUTION_NAME` and improves tracking of installed shortcuts.

### Changes

- Add `MENUINST_DISTRIBUTION_NAME` env var support with resolution order:
  1. `MENUINST_DISTRIBUTION_NAME` env var (highest priority)
  2. `$BASE_PREFIX/Menu/menuinst.toml` value
  3. `base_prefix.name` (fallback)
- Add `menuinst.toml` for persistent storage of `distribution_name` and shortcut tracking
- Add `read_menuinst_toml()` in `utils.py` and `write_menuinst_toml()` in `api.py`
- Add `record_shortcuts()` / `remove_shortcut_records()` called from `install()` / `remove()`
- Add `tomli` (py<3.11) and `tomli-w` dependencies

### Planned changes once on main
1. constructor: Set `MENUINST_DISTRIBUTION_NAME=%APPNAME%` in `run_installation.bat` for MSI installers
2. Integration tests - Add tests that call `install()`/`remove()` with real menu JSON to verify end-to-end TOML recording
3. Use recorded paths for removal - Update `remove()` to use the recorded shortcut paths from `menuinst.toml` whenever possible instead of recomputing from menu JSON metadata

### Checklist - did you ...

### DIFF
--- a/menuinst/api.py
+++ b/menuinst/api.py
@@ -283,7 +283,8 @@ def _install_adapter(path: str, remove: bool = False, prefix: str = DEFAULT_PREF
         kwargs.setdefault("base_prefix", kwargs.pop("root_prefix", DEFAULT_BASE_PREFIX))
         if kwargs["base_prefix"] is None:
             kwargs["base_prefix"] = DEFAULT_BASE_PREFIX
+        # Pass path so install/remove records the actual filename in menuinst.toml
         if remove:
-            _api_remove(metadata, target_prefix=prefix, **kwargs)
+            _api_remove(json_path, target_prefix=prefix, **kwargs)
         else:
-            install(metadata, target_prefix=prefix, **kwargs)
+            install(json_path, target_prefix=prefix, **kwargs)
--- a/menuinst/api.py
+++ b/menuinst/api.py
@@ -10,17 +10,15 @@ from logging import getLogger
 from pathlib import Path
 from typing import Callable, Union
 
-import tomli_w
-
 from .platforms import Menu, MenuItem
 from .utils import (
     DEFAULT_BASE_PREFIX,
     DEFAULT_PREFIX,
-    MENUINST_TOML_SCHEMA_VERSION,
     _UserOrSystem,
     elevate_as_needed,
     read_menuinst_toml,
     user_is_admin,
+    write_menuinst_toml,
 )
 
 log = getLogger(__name__)
@@ -40,18 +38,6 @@ def _maybe_try_user(base_prefix: str, target_prefix: str) -> bool:
     if Path(target_prefix, ".nonadmin").is_file():
         return True
     return Path(base_prefix, ".nonadmin").is_file()
-
-
-def write_menuinst_toml(prefix: Path, data: dict) -> None:
-    """Write menuinst.toml atomically."""
-    data.setdefault("schema_version", MENUINST_TOML_SCHEMA_VERSION)
-    menu_dir = prefix / "Menu"
-    menu_dir.mkdir(parents=True, exist_ok=True)
-    toml_path = menu_dir / "menuinst.toml"
-    tmp_path = menu_dir / "menuinst.toml.tmp"
-    with open(tmp_path, "wb") as f:
-        tomli_w.dump(data, f)
-    tmp_path.replace(toml_path)
 
 
 def record_shortcuts(
--- a/menuinst/api.py
+++ b/menuinst/api.py
@@ -10,6 +10,12 @@ from logging import getLogger
 from pathlib import Path
 from typing import Callable, Union
 
+try:
+    import tomllib
+except ImportError:
+    import tomli as tomllib
+import tomli_w
+
 from .platforms import Menu, MenuItem
 from .utils import (
     DEFAULT_BASE_PREFIX,
@@ -20,6 +26,7 @@ from .utils import (
 )
 
 log = getLogger(__name__)
+MENUINST_TOML_SCHEMA_VERSION = 1
 
 
 __all__ = [
@@ -36,6 +43,82 @@ def _maybe_try_user(base_prefix: str, target_prefix: str) -> bool:
     if Path(target_prefix, ".nonadmin").is_file():
         return True
     return Path(base_prefix, ".nonadmin").is_file()
+
+
+def read_menuinst_toml(prefix: Path) -> dict:
+    """Read menuinst.toml from prefix, returning empty dict if missing/invalid."""
+    toml_path = prefix / "Menu" / "menuinst.toml"
+    if not toml_path.exists():
+        return {}
+    try:
+        with open(toml_path, "rb") as f:
+            data = tomllib.load(f)
+        version = data.get("schema_version", 1)
+        if version > MENUINST_TOML_SCHEMA_VERSION:
+            log.warning(
+                "menuinst.toml at %s has schema_version %d, "
+                "but this menuinst only understands version %d",
+                toml_path,
+                version,
+                MENUINST_TOML_SCHEMA_VERSION,
+            )
+        return data
+    except Exception as e:
+        log.warning("Failed to read menuinst.toml from %s: %s", toml_path, e)
+        return {}
+
+
+def write_menuinst_toml(prefix: Path, data: dict) -> None:
+    """Write menuinst.toml atomically."""
+    data.setdefault("schema_version", MENUINST_TOML_SCHEMA_VERSION)
+    menu_dir = prefix / "Menu"
+    menu_dir.mkdir(parents=True, exist_ok=True)
+    toml_path = menu_dir / "menuinst.toml"
+    tmp_path = menu_dir / "menuinst.toml.tmp"
+    with open(tmp_path, "wb") as f:
+        tomli_w.dump(data, f)
+    tmp_path.replace(toml_path)
+
+
+def record_shortcuts(
+    prefix: Path,
+    base_prefix: Path,
+    source: str,
+    paths: list[os.PathLike],
+    distribution_name: str | None = None,
+) -> None:
+    """Record created shortcuts to menuinst.toml."""
+    if not paths:
+        return
+
+    data = read_menuinst_toml(prefix)
+
+    # Write distribution_name only to base prefix, and only if not already set
+    if prefix.samefile(base_prefix) and distribution_name:
+        data.setdefault("distribution_name", distribution_name)
+
+    # Append shortcuts
+    shortcuts = data.setdefault("shortcuts", [])
+    for path in paths:
+        shortcuts.append({"source": source, "path": str(path)})
+
+    write_menuinst_toml(prefix, data)
+
+
+def remove_shortcut_records(prefix: Path, source: str) -> None:
+    """Remove shortcut entries matching source from menuinst.toml."""
+    data = read_menuinst_toml(prefix)
+    if not data:
+        return
+
+    shortcuts = data.get("shortcuts", [])
+    if not shortcuts:
+        return
+
+    # Filter out entries matching this source
+    data["shortcuts"] = [s for s in shortcuts if s.get("source") != source]
+
+    write_menuinst_toml(prefix, data)
 
 
 def _load(
@@ -77,6 +160,19 @@ def install(
     for menu_item in menu_items:
         paths += menu_item.create()
 
+    # Record shortcuts to menuinst.toml
+    if isinstance(metadata_or_path, (str, Path)):
+        source = Path(metadata_or_path).name
+    else:
+        source = f"{menu.name}.json"
+    record_shortcuts(
+        Path(target_prefix),
+        Path(base_prefix),
+        source,
+        paths,
+        distribution_name=menu.placeholders.get("DISTRIBUTION_NAME"),
+    )
+
     return paths
 
 
@@ -107,6 +203,13 @@ def remove(
         for menu_item in menu_items:
             paths += menu_item.remove()
         paths += menu.remove()
+
+    # Remove shortcut records from menuinst.toml
+    if isinstance(metadata_or_path, (str, Path)):
+        source = Path(metadata_or_path).name
+    else:
+        source = f"{menu.name}.json"
+    remove_shortcut_records(Path(target_prefix), source)
 
     return paths
 
--- a/menuinst/api.py
+++ b/menuinst/api.py
@@ -80,7 +80,12 @@ def record_shortcuts(
 
 
 def remove_shortcut_records(prefix: Path, source: str) -> None:
-    """Remove shortcut entries matching source from menuinst.toml."""
+    """Remove shortcut entries matching source from menuinst.toml.
+
+    TODO: Use the recorded paths as the source of truth for shortcut removal,
+    instead of recomputing paths from menu JSON metadata. This would handle
+    cases where shortcuts were moved or the menu JSON changed.
+    """
     data = read_menuinst_toml(prefix)
     if not data:
         return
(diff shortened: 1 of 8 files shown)


menuinst/api.py
```diff
@@ -198,7 +283,8 @@ def _install_adapter(path: str, remove: bool = False, prefix: str = DEFAULT_PREF
         kwargs.setdefault("base_prefix", kwargs.pop("root_prefix", DEFAULT_BASE_PREFIX))
         if kwargs["base_prefix"] is None:
             kwargs["base_prefix"] = DEFAULT_BASE_PREFIX
+        # Pass path so install/remove records the actual filename in menuinst.toml
         if remove:
-            _api_remove(metadata, target_prefix=prefix, **kwargs)
+            _api_remove(json_path, target_prefix=prefix, **kwargs)
```
lrandersson, line 274: This change was done to properly register the file name the shortcuts originate from in the toml file. There are other ways to achieve the same behavior but this approach required the least intrusive code changes.

menuinst/cli/cli.py
```diff
@@ -1,11 +1,13 @@
...
 
 import argparse
+import os
 import re
 import sys
 from pathlib import Path
 
-from ..api import _install_adapter
+from ..api import _install_adapter, write_menuinst_toml
+from ..utils import read_menuinst_toml
```
marcoesters: I think I understand why you're doing this ("public" API vs "private" utility function), but it looks awkward to have the read and write functions in different modules. I'm not sure if we should add the read function to the API just because of that though.

menuinst/cli/cli.py
```diff
@@ -70,6 +72,17 @@ def install(
     if root_prefix:
         root_prefix = str(Path(root_prefix).expanduser().resolve())
 
+    # Persist distribution_name from env var before processing shortcuts.
+    # The env var allows installers to set distribution_name dynamically at
+    # install time. We must persist it here because the env var is transient
+    # and may not be set when packages with shortcuts are installed later.
+    if env_name := os.environ.get("MENUINST_DISTRIBUTION_NAME"):
```
marcoesters: Let's call it `distribution_name` to be consistent.

I'm wondering though if it makes sense to record all `MENUINST_*` environment variables here.

I'm just generally wondering how we would distinguish between those "global" parameters that should persist vs. parameters that should be recalculated during install-time. Maybe we need a separate command for that? `menuinst --set-global-placeholder` or something similar?

menuinst/platforms/base.py
```diff
@@ -74,7 +91,7 @@ def placeholders(self) -> dict[str, str]:
         """
         return {
             "BASE_PREFIX": str(self.base_prefix),
-            "DISTRIBUTION_NAME": self.base_prefix.name,
+            "DISTRIBUTION_NAME": self._get_distribution_name(),
             "BASE_PYTHON": str(self.base_prefix / "bin" / "python"),
```
marcoesters, line 95: Since we already have `MENUINST_BASE_PREFIX` and the like, I wonder if we can generalize overwriting those placeholders instead of special-casing `DISTRIBUTION_NAME` (even though `BASE_PREFIX` and `PREFIX` are special cases).

menuinst/api.py
```diff
@@ -38,6 +42,67 @@ def _maybe_try_user(base_prefix: str, target_prefix: str) -> bool:
...
+) -> None:
+    """Record created shortcuts to menuinst.toml."""
+    if not paths:
+        return
+
+    data = read_menuinst_toml(prefix)
+
+    # Write distribution_name only to base prefix, and only if not already set
+    if prefix.samefile(base_prefix) and distribution_name:
+        data.setdefault("distribution_name", distribution_name)
```
marcoesters, line 58: I think this illustrates the brittleness of the special casing a little bit. If we come up with other variables that must only be written into the base prefix, we have to keep making sure that they are in the correct place. Using the `menuinst --set-global-placeholder` would make make this more stable.

menuinst/utils.py
```diff
@@ -13,7 +13,13 @@
 from typing import Any, Callable, Iterable, Literal, Mapping, Optional, Sequence, Union
 from unicodedata import normalize
 
+try:
+    import tomllib
+except ImportError:
+    import tomli as tomllib
+
 logger = getLogger(__name__)
+MENUINST_TOML_SCHEMA_VERSION = 1
```
marcoesters: I suspect we will need to be more fine-grained than that?

menuinst/utils.py
```diff
@@ -70,6 +76,29 @@ def _default_prefix(which: _TargetOrBase = "target") -> str:
...
+                "menuinst.toml at %s has schema_version %d, "
+                "but this menuinst only understands version %d",
+                toml_path,
+                version,
+                MENUINST_TOML_SCHEMA_VERSION,
+            )
+        return data
+    except Exception as e:
+        logger.warning("Failed to read menuinst.toml from %s: %s", toml_path, e)
+        return {}
```
marcoesters: I think we should actually error out here. The only way I can reasonably imagine that the file is invalid is if someone manipulated the file manually. In that case, returning `{}` may result in unexpected behavior.

menuinst/api.py
```diff
@@ -38,6 +42,67 @@ def _maybe_try_user(base_prefix: str, target_prefix: str) -> bool:
...
+
+    write_menuinst_toml(prefix, data)
+
+
+def remove_shortcut_records(prefix: Path, source: str) -> None:
+    """Remove shortcut entries matching source from menuinst.toml.
+
+    TODO: Use the recorded paths as the source of truth for shortcut removal,
+    instead of recomputing paths from menu JSON metadata. This would handle
+    cases where shortcuts were moved or the menu JSON changed.
```
marcoesters, line 73: I may be missing some of the complexity here, but I think this is well in scope of this PR. All the pieces are already there.

tests/test_cli.py
```diff
@@ -114,3 +111,33 @@ def test_cli_errors(argv: list[str]) -> None:
...
+    "env_var_set,expect_toml",
+    [
+        pytest.param(True, True, id="env_var_set_creates_toml"),
+        pytest.param(False, False, id="no_env_var_no_toml"),
+    ],
+)
+def test_cli_env_var_persistence(
+    tmp_path: Path, monkeypatch: pytest.MonkeyPatch, env_var_set: bool, expect_toml: bool
+) -> None:
+    """Test that MENUINST_DISTRIBUTION_NAME env var is persisted to menuinst.toml."""
```
marcoesters:
```suggestion
    """Test that MENUINST_DISTRIBUTION_NAME env var persists in menuinst.toml."""
```

marcoesters (commented):
This is nice - I am a little worried about the special-casing of `DISTRIBUTION_NAME` and made an alternative suggestion. I haven't thought through all the ramifications of that suggestion yet though.

menuinst/cli/cli.py
lrandersson: Yeah the intent was that the utils is read-only but the other ones actually cause changes. I originally had them all n `api.py` but decided last minute to move it. I'm fine with either. Proposals: 
  1. Move both to utils.py (keep read/write together)
  2. Move both to api.py (where mutations happen)
  3. Keep as-is and document the rationale

menuinst/cli/cli.py
lrandersson: I think the generalization is interesting but comes with additional complexity, which I would propose we postpone until we have concrete use cases. Refactoring this wouldn't be difficult but I'm afraid we run into fixing something where we don't have any real use case (if that makes sense).
Also renamed the variable: d5324fb

menuinst/platforms/base.py
lrandersson: Yeah my answer to this is similar to above. I think the idea is good but something I think we can do better once we have concrete use cases. I would also argue that `MENUINST_BASE_PREFIX` and `MENUINST_PREFIX` are runtime path overrides and are slightly different from `DISTRIBUTION_NAME` in this scenario.

menuinst/api.py
lrandersson: Yeah I think that's definitely a valid concern. I think in its current state with only one variable, it's doable, but if we add more I think definitely we can iterate on the design to something more robust. I've been trying to keep it simple and avoid over-engineering.

menuinst/api.py
lrandersson: I added this TODO to help the reviewers and keep it simple. Merging this wouldn't break any existing behavior. I'm happy to do it in this PR but perhaps even opening a separate PR targeting this branch would be more feasible, let me know what you prefer.

menuinst/utils.py
lrandersson: You mean something like `1.0.0`? Absolutely, I just made it simple (`1`) right now to illustrate the purpose.

menuinst/utils.py
lrandersson: See d5324fb

tests/test_cli.py
lrandersson: d5324fb

menuinst/utils.py
marcoesters: I would use SchemaVer like we do for the `menuinst` schema: https://docs.snowplow.io/docs/api-reference/iglu/common-architecture/schemaver/

menuinst/cli/cli.py
marcoesters: I don't have too strong of an opinion either way, but 2. for the "main" TOML functions (top-level read and write) sounds the most sensible to me.

menuinst/cli/cli.py
marcoesters: I was thinking about #334 in particular. Or the [environment placeholder](https://github.com/AnacondaRecipes/anaconda_powershell_prompt-feedstock/blob/dfb1fdb00f0f2b5d3e28138ac1288ca3d7aee673/recipe/menu.json#L6) in the Anaconda Prompt menus or the [package version placeholders](https://github.com/conda-forge/spyder-feedstock/blob/9b91895279cb02e92ac6e71e600f672934909f75/recipe/spyder-menu-win.json#L8-L9) for `spyder`. Right now, this needs to be implemented with a [`sed`-like workaround](https://github.com/conda-forge/spyder-feedstock/blob/9b91895279cb02e92ac6e71e600f672934909f75/recipe/post-link.bat#L69-L79) for Windows.

Setting variables like this via `menuinst` and then using placeholder syntax would be really convenient and remove all special casing. Installers would then just call `_conda.exe menuinst --set-(global|local)-placeholder CUSTOM_PLACEHOLDER=something` (or similar).

menuinst/api.py
marcoesters: A separate PR targeting this branch would work for me. As of this PR, all we do with those recorded paths is remove them from the record on uninstallation. So, the records currently don't find much use.

lrandersson:
@marcoesters for the second round of review, see my responses below (trying to avoid spamming with email notifications):
1. Use `SchemaVer` format: df00f53
2. Move functions such that read/write are now in the same module (both in `utils.py`): df00f53
3. Regarding the custom placeholders: https://github.com/conda/menuinst/pull/477#discussion_r3241873748
I see the value and agree that it would be good to add, but to reduce scope creep of this PR which is intended primarily to resolve the MSI installer issue I propose that I create a separate issue as a follow-up to this one, since as far as I understand there are questions that we need to address that I don't think is ideal to resolve via review comments:
  - **CLI interface design** - `--set-global-placeholder KEY=value` vs `--set-local-placeholder KEY=value`
  - **Storage format** - How to distinguish global vs local placeholders in `menuinst.toml`? As an example: Currently menuinst.toml has distribution_name as a top-level key. For arbitrary placeholders, we need to decide:
    - Maybe a flat structure? `FOO = "value"` at root level (risk: collision with future schema keys)
    - or nested under a section? `[placeholders]` section with `FOO = "value"`
  - **Resolution ordering** - When resolving `{{ FOO }}` in a menu.json, what takes precedence?
    - Environment variable `MENUINST_FOO`?
    - Value in current prefix's menuinst.toml?
    - Value in base_prefix's menuinst.toml?
  - **Scope semantics** - global = base_prefix only vs local = per-environment?
Let me know what you think, just to recall, I think the idea is great but I just think it should be done as part of a separate PR.

menuinst/api.py
lrandersson: @marcoesters please see this separate PR https://github.com/lrandersson/menuinst/pull/1 targeting this functionality.

marcoesters:
> @marcoesters for the second round of review, see my responses below (trying to avoid spamming with email notifications):
> 
>     1. Use `SchemaVer` format: [df00f53](https://github.com/conda/menuinst/commit/df00f5312adf230cfec53710857ac588d865bf23)
> 
>     2. Move functions such that read/write are now in the same module (both in `utils.py`): [df00f53](https://github.com/conda/menuinst/commit/df00f5312adf230cfec53710857ac588d865bf23)
> 
>     3. Regarding the custom placeholders: [Add support for menuinst.toml #477 (comment)](https://github.com/conda/menuinst/pull/477#discussion_r3241873748)
>        I see the value and agree that it would be good to add, but to reduce scope creep of this PR which is intended primarily to resolve the MSI installer issue I propose that I create a separate issue as a follow-up to this one, since as far as I understand there are questions that we need to address that I don't think is ideal to resolve via review comments:
> 
> 
>     * **CLI interface design** - `--set-global-placeholder KEY=value` vs `--set-local-placeholder KEY=value`
> 
>     * **Storage format** - How to distinguish global vs local placeholders in `menuinst.toml`? As an example: Currently menuinst.toml has distribution_name as a top-level key. For arbitrary placeholders, we need to decide:
>       
>       * Maybe a flat structure? `FOO = "value"` at root level (risk: collision with future schema keys)
>       * or nested under a section? `[placeholders]` section with `FOO = "value"`
> 
>     * **Resolution ordering** - When resolving `{{ FOO }}` in a menu.json, what takes precedence?
>       
>       * Environment variable `MENUINST_FOO`?
>       * Value in current prefix's menuinst.toml?
>       * Value in base_prefix's menuinst.toml?
> 
>     * **Scope semantics** - global = base_prefix only vs local = per-environment?
>       Let me know what you think, just to recall, I think the idea is great but I just think it should be done as part of a separate PR.

These are all valid questions - I'm okay with punting that to a different issue.

menuinst/api.py
marcoesters: Turns out I did miss the complexity. 😅

tests/test_cli.py
```diff
@@ -1,17 +1,14 @@
 from __future__ import annotations
 
 import json
-from typing import TYPE_CHECKING
+from pathlib import Path
```
marcoesters: I think this needs to go back into the `TYPE_CHECKING` block.

tests/test_metadata.py
```diff
@@ -0,0 +1,221 @@
...
+
+# Placeholder distribution names for tests
+DIST_NAME = "Something"
+DIST_NAME_ALT = "SomethingElse"
+
+
+class TestGetDistributionName:
+    """Tests for Menu._get_distribution_name() resolution order."""
+
+    def test_env_var_takes_priority(self, tmp_path, monkeypatch):
```
marcoesters: The input parameters are missing types.

marcoesters (commented):
Just minor typing comments, but LGTM otherwise.

tests/test_metadata.py
lrandersson: See 63e3fb0

tests/test_cli.py
lrandersson: See 72fe87d

marcoesters (approved):

Issue #496, opened 2026-06-04
Test failures in Fedora package


opoplawski:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

I'm updating the Fedora python-menuinst package to 2.5.0 and getting the following test failures:
```
+ /usr/bin/pytest -rs --ignore=tests/test_elevation.py
============================= test session starts ==============================
platform linux -- Python 3.14.5, pytest-9.0.3, pluggy-1.6.0
rootdir: /builddir/build/BUILD/python-menuinst-2.5.0-build/menuinst-2.5.0
configfile: pyproject.toml
collected 90 items / 4 skipped
tests/_legacy/test_menu_creation.py ssss                                 [  4%]
tests/_legacy/test_win32.py s                                            [  5%]
tests/test_api.py F..FFFsssssssss.......                                 [ 30%]
tests/test_cli.py ......                                                 [ 36%]
tests/test_conda.py ..ss                                                 [ 41%]
tests/test_conda_plugin.py ......                                        [ 47%]
tests/test_data.py ...                                                   [ 51%]
tests/test_metadata.py ....................                              [ 73%]
tests/test_schema.py ..............                                      [ 88%]
tests/test_utils.py ssssssssss                                           [100%]
=================================== FAILURES ===================================
_____________________________ test_install_prefix ______________________________
delete_files = ['/tmp/tmpgdi12qmysys-prefix.json', '/tmp/tmpx_flo85z']
    def test_install_prefix(delete_files):
>       check_output_from_shortcut(delete_files, "sys-prefix.json", expected_output=sys.prefix)
tests/test_api.py:146: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/test_api.py:66: in check_output_from_shortcut
    paths = install(abs_json_path, base_prefix=tmp_base_path)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../BUILDROOT/usr/lib/python3.14/site-packages/menuinst/utils.py:588: in wrapper_elevate
    return func(
../BUILDROOT/usr/lib/python3.14/site-packages/menuinst/api.py:136: in install
    record_shortcuts(
[...]

PR #497, opened 2026-06-05
Make tests more robust and gracefully handle permission denied when creating `menuinst.toml` 

lrandersson:
### Description

This PR is created based on the feedback in https://github.com/conda/menuinst/issues/496.

When creating the feature for `menuinst.toml` we didn't account for `target_prefix` being read only, however, I believe that the origin of the new feature (https://github.com/conda/menuinst/pull/477) isn't the root cause of the issue but in fact that this revealed a previously hidden issue in the tests - in particular we shouldn't try to create files in `/usr` during the testing.

Tests now use a `temp` directory with a symlink/copy of` Python` instead of relying on `sys.prefix`, in particular since before these changes we relied on `/usr` and extrapolated this to `/usr/bin/python`.

#### Changes:

1. `check_output_from_shortcut` refactor — Use `temp` directory as `target_prefix` instead of `sys.prefix`. Create a symlink (Unix) or copy (Windows) of the Python executable so `{{ PYTHON }}` placeholder resolves correctly.
2. macOS test fixes — Several tests had latent bugs where `remove()` was called without prefixes. Now properly capture and pass `target_prefix`/`base_prefix`.
3. `PermissionError` handling — `record_shortcuts()` and `remove_shortcut_records()` now gracefully handle permission errors with a debug log instead of crashing. This also seems to be similar with how `.nonadmin` is handled. This allows `menuinst` to work in read-only environments (shortcut tracking is skipped, but installation succeeds).

### Checklist - did you ...

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -77,11 +77,18 @@ def check_output_from_shortcut(
         # We symlink to sys.executable; copying doesn't work (can't find runtime libraries).
         if PLATFORM == "win":
             python_path = Path(tmp_base_path) / "python.exe"
+            try:
+                python_path.symlink_to(sys.executable)
+            except OSError:
+                # If symlinks are not available;
+                # fall back to sys.prefix; tests lose isolation but still run.
+                delete_files.remove(tmp_base_path)
+                tmp_base_path = sys.prefix
         else:
             bin_dir = Path(tmp_base_path) / "bin"
             bin_dir.mkdir(exist_ok=True)
             python_path = bin_dir / "python"
-        python_path.symlink_to(sys.executable)
+            python_path.symlink_to(sys.executable)
 
     paths = install(abs_json_path, target_prefix=tmp_base_path, base_prefix=tmp_base_path)
     try:
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -63,7 +63,7 @@ def check_output_from_shortcut(
     tmp_base_path = mkdtemp()
     delete_files.append(tmp_base_path)
     (Path(tmp_base_path) / ".nonadmin").touch()
-    paths = install(abs_json_path, base_prefix=tmp_base_path)
+    paths = install(abs_json_path, target_prefix=sys.prefix, base_prefix=tmp_base_path)
     try:
         if action == "run_shortcut":
             if PLATFORM == "win":
@@ -114,7 +114,7 @@ def check_output_from_shortcut(
         if paths:
             delete_files += list(paths)
         if remove_after:
-            remove(abs_json_path, base_prefix=tmp_base_path)
+            remove(abs_json_path, target_prefix=sys.prefix, base_prefix=tmp_base_path)
         if PLATFORM == "osx" and action in ("open_file", "open_url"):
             _lsregister(
                 "-kill",
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -73,7 +73,8 @@ def check_output_from_shortcut(
         (Path(tmp_base_path) / ".nonadmin").touch()
         # conda-meta makes conda treat this as a valid environment for activation
         (Path(tmp_base_path) / "conda-meta").mkdir(exist_ok=True)
-        # Symlink to real Python so {{ PYTHON }} placeholder works
+        # Shortcuts use {{ PYTHON }} which resolves to prefix/python.exe or prefix/bin/python.
+        # We symlink to sys.executable; copying doesn't work (can't find runtime libraries).
         if PLATFORM == "win":
             python_path = Path(tmp_base_path) / "python.exe"
         else:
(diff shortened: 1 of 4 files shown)


tests/test_metadata.py
```diff
@@ -155,6 +158,63 @@ def test_distribution_name_only_written_to_base_prefix(self, tmp_path: Path) ->
         assert "distribution_name" not in data
         assert len(data["shortcuts"]) == 1
 
+    @pytest.mark.skipif(sys.platform == "win32", reason="chmod doesn't work on Windows")
```
marcoesters: We could use `icacls.exe` or the Windows API to change the permissions.

tests/test_api.py
```diff
@@ -60,10 +60,29 @@ def check_output_from_shortcut(
         abs_json_path = tmp.name
         delete_files.append(abs_json_path)
 
-    tmp_base_path = mkdtemp()
-    delete_files.append(tmp_base_path)
-    (Path(tmp_base_path) / ".nonadmin").touch()
-    paths = install(abs_json_path, base_prefix=tmp_base_path)
+    # Windows file/URL association tests need sys.prefix because symlinked Python
```
marcoesters, line 63: Could we just copy the Python executable on Windows? Symlinks aren't always allowed on Windows anyway.

tests/test_api.py
```diff
@@ -60,10 +60,29 @@ def check_output_from_shortcut(
...
+
+    if use_real_prefix:
+        tmp_base_path = sys.prefix
+    else:
+        tmp_base_path = mkdtemp()
+        delete_files.append(tmp_base_path)
+        (Path(tmp_base_path) / ".nonadmin").touch()
+        # conda-meta makes conda treat this as a valid environment for activation
+        (Path(tmp_base_path) / "conda-meta").mkdir(exist_ok=True)
+        # Symlink to real Python so {{ PYTHON }} placeholder works
```
marcoesters: Can you elaborate on why the Python executable must be in the target/base prefix?

lrandersson:
@marcoesters I tried to address all the review comments in the same commit: 30e7f963651b08e5cbc4ba2b25c9d3e55afd5c9a

tests/test_api.py
```diff
@@ -60,10 +60,30 @@ def check_output_from_shortcut(
...
+        (Path(tmp_base_path) / "conda-meta").mkdir(exist_ok=True)
+        # Shortcuts use {{ PYTHON }} which resolves to prefix/python.exe or prefix/bin/python.
+        # We symlink to sys.executable; copying doesn't work (can't find runtime libraries).
+        if PLATFORM == "win":
+            python_path = Path(tmp_base_path) / "python.exe"
+        else:
+            bin_dir = Path(tmp_base_path) / "bin"
+            bin_dir.mkdir(exist_ok=True)
+            python_path = bin_dir / "python"
+        python_path.symlink_to(sys.executable)
```
marcoesters: This will fail on Windows if symlinks are not enabled (and symlinks are not enabled by default).

lrandersson: That's a good point. I wonder if we should resort to skipping the tests if Windows symlinks are unavailable. Do you have any other proposals? This was the only approach I managed to get working so far.

marcoesters: I'm not sure about that. We would essentially skip all those Windows tests now that used to be covered.

We are just linking here because Python does not exist in `prefix`, right? What are we actually defending ourself against with these Python changes? It seems to me like the tests have worked the way they needed to before and only failed because of the write protection. Your new test in `test_metadata.py` ensures that this doesn't happen anymore.

Maybe we just need to add (instead of what we have in the metadata tests?) a test that we can install `sys-prefix.json` from a read-only location and leave the `check_output_from_shortcut` the way it was before?

opoplawski:
This appears to fix #496 , Thanks!

tests/test_api.py
lrandersson: - Correct we introduced the `Python` simlink since, after changing to `/tmp` directories, the path `/tmp/<something>/bin/python` does not exist, and these tests verify that the shortcut actually runs (which is why it needs Python in the first place).
- I see your point to not do the refactoring, however, I still find it odd that the tests are setup such that we potentially are creating files in `/usr`. There are already many tests that leave files on disk but ideally I'd like all the tests to be designed such that we do not affect the local file system (i.e. a developer should be able to run all the unit tests locally without cluttering the file system).
That being said, **if** we go with that approach (to answer the original question), perhaps we could go with the following approach:
* We try symlink, and if that doesn't work we fall back to `sys.prefix`. This means we do have test isolation in most cases (but not when symlinks are unavailable). As it is today; 4 tests are currently affected by this (which isn't too many).

marcoesters: > We try symlink, and if that doesn't work we fall back to sys.prefix.

That seems to be more robust.

lrandersson: See 334ca5f

marcoesters (approved):

PR #535, opened 2026-08-22
Fix missing f-string prefix on the Icon= desktop entry line

hmaarrfk:
I verified the AI's work to the best of my knowledge. seems sound.

> **Opened by an AI agent (Claude) acting on behalf of @hmaarrfk, who has reviewed and takes responsibility for this change.**

### Description

`LinuxMenuItem._write_desktop_file` writes the `Icon=` line as a plain string literal rather than an f-string, so every generated `.desktop` file gets the literal source text as its icon value:

```python
lines.append("Icon={_escape_desktop_string(self.render_key('icon')}")
```

The brace is also unbalanced (`self.render_key('icon')` never closed), which is further evidence the expression was never evaluated.

Result, reproduced on `main`:

```
[Desktop Entry]
Type=Application
Encoding=UTF-8
Name=Test Item
Exec=bash -c '... && echo hi'
Terminal=false
Icon={_escape_desktop_string(self.render_key('icon')}
```

Every shortcut with an `icon` ships with a broken `Icon=` value, so desktops fall back to a generic icon.

#### Provenance

This is a regression. It was introduced in #515 (`9efbfbd0`, "Fix content writer"), which converted the writer from a `configparser` dictionary into a list of lines:

```diff
-  config["Desktop Entry"]["Icon"] = _escape_desktop_string(self.render_key("icon"))
+  lines.append("Icon={_escape_desktop_string(self.render_key('icon')}")
```

The sibling `Exec=` line in that same commit *did* get its `f` prefix, so this looks like an oversight on one line rather than an intentional change. `2.5.2` is the only release carrying it so far; `2.4.2` and earlier are unaffected.

#### The patch

Restores the `f` prefix and the missing closing brace — one character each. Behaviour matches the pre-#515 `configparser` version exactly.

#### Testing

Adds `tests/test_desktop_entry_icon`, which writes a desktop entry for an item with an `icon` and asserts the `Icon=` line holds the rendered path. It exercises `_write_desktop_file` directly rather than going through `install()`, so it runs on every platform rather than needing a Linux desktop session.

Verified it is a real regression test:

```
$ pytest tests/test_api.py::test_desktop_entry_icon   # without the one-line fix
FAILED tests/test_api.py::test_desktop_entry_icon - assert 'Icon=<prefix>/Menu/icon.png' in [...]
1 failed

$ pytest tests/test_api.py::test_desktop_entry_icon   # with the fix
1 passed
```

Full suite on macOS (Python 3.12): `74 passed, 27 skipped, 1 failed`. The single failure is `tests/test_elevation.py::test_elevation`, which fails identically on unmodified `main` in this environment and is unrelated. I could not run the Linux-only tests (`test_desktop_files_escaping` and the other `PLATFORM == "linux"` cases) — no Linux desktop session available here; CI covers those.

`pre-commit run --all-files` passes (all 19 hooks).

### Checklist - did you ...

- [ ] ~Add / update outdated documentation?~ No user-facing documentation describes this line.

### DIFF
--- a/menuinst/platforms/linux.py
+++ b/menuinst/platforms/linux.py
@@ -264,7 +264,7 @@ class LinuxMenuItem(MenuItem):
 
         icon = self.render_key("icon")
         if icon:
-            lines.append("Icon={_escape_desktop_string(self.render_key('icon')}")
+            lines.append(f"Icon={_escape_desktop_string(self.render_key('icon'))}")
 
         description = self.render_key("description")
         if description:
--- a/news/535-fix-desktop-entry-icon
+++ b/news/535-fix-desktop-entry-icon
@@ -1,0 +1,21 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Write the rendered icon path to the `Icon=` line of Linux `.desktop` files.
+  A missing f-string prefix caused the literal source expression to be written
+  instead. (#535)
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* <news item>
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -632,6 +632,7 @@ def test_desktop_files_escaping(delete_files):
     check_output_from_shortcut(delete_files, "pwnd.json", expected_output="legit")
 
 
+@pytest.mark.skipif(PLATFORM != "linux", reason="Only relevant to .desktop files")
 def test_desktop_entry_icon(tmp_path, monkeypatch):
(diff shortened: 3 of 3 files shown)


lrandersson:
Thank you @hmaarrfk for submitting this. I just approved the workflow for the tests, I think the changes and summary looks good. I'll approve it once the tests are green.

hmaarrfk:
thanks i added the skip to the test.

marcoesters (approved):

hmaarrfk:
Thank you!

PR #553, opened 2026-09-15
Simplify conda activation behavior in Windows shortcuts

### LINKED ISSUE #321: CMD shell activation changed in `conda 25.3.0`

### LINKED ISSUE #468: Convoluted windows shortcut code causing security problems
### Checklist

### What happened?

When creating a shortcut on Windows, the code that the shortcut is running is something like this:

```
C:\WINDOWS\system32\cmd.exe /D /C START /MIN "" "C:\WINDOWS\system32\WindowsPowerShell\v1.0\powershell.exe" -WindowStyle hidden "start 'C:\Users\path\to\miniforge3\envs\myapp_env\Menu\myapp 1.0.0.bat' -WindowStyle hidden"
```

Something about this is being flagged by our security software as malicious. Why is it so convoluted? Is there an easier way than having one shell calling another shell and having multiple hidden flags in the mix?

### LINKED ISSUE #545: Windows shortcuts can silently omit CONDA_PREFIX / env config vars from the activation script
### Checklist

### What happened?

**Component:** `menuinst/platforms/win.py`, `WindowsMenuItem._command()` (the `.env`-file branch used for conda >= 25.3.0)

### Summary

When a package's post-install hook calls `menuinst.api.install()` from *inside* the environment it's creating a shortcut for (i.e. `CONDA_SHLVL`/`CONDA_PREFIX`/`CONDA_DEFAULT_ENV` already point at that same target prefix), the generated `.bat` shortcut is missing `CONDA_PREFIX`, `CONDA_DEFAULT_ENV`, and any custom variables set via `conda env config vars set` (from `conda-meta/state`). `PATH` and the `activate.d`/`deactivate.d` hook calls are present and correct, so the shortcut looks activated but isn't fully.

### Root cause

`_command()` shells out to `conda shell.cmd.exe activate <prefix>` via `logged_run(activator_cmd, check=True, log=False)` with no explicit `env=`, so the subprocess inherits the calling process's environment as-is. `conda`'s activator only emits `SET` lines for variables whose value differs from what's already in that inherited environment. If the builder process is already activated in the same target env, `CONDA_PREFIX`/`CONDA_DEFAULT_ENV`/config vars already match the target exactly, so conda considers them "unchanged" and omits them — even though the resulting script is meant to run standalone, later, in a fresh `cmd.exe` with none of that ambient state.

### Reproduction

1. Env `myenv` has a custom var set: `conda env config vars set FOO=bar -n myenv`.
2. From a **clean** shell: `conda shell.cmd.exe activate <prefix of myenv>` → resulting `.env` file includes `CONDA_PREFIX`, `CONDA_DEFAULT_ENV`, and `FOO=bar`.
3. From a shell where `myenv` is **already active** (`conda activate myenv` first): re-run the same `shell.cmd.exe activate <prefix>` call → the `.env` file now omits `CONDA_PREFIX`, `CONDA_DEFAULT_ENV`, and `FOO` entirely, while still including `PATH`, `CONDA_SHLVL`, and the `activate.d`/`deactivate.d` `_CONDA_SCRIPT` lines.
4. Since `menuinst`'s post-install-triggered shortcut generation typically requires being run activated (to know `CONDA_PREFIX` in the first place), any package following that pattern gets a shortcut baked from case 3, not case 2.

### Expected

The generated shortcut should be self-contained and always include the full activation state (`CONDA_PREFIX`, `CONDA_DEFAULT_ENV`, and all `conda-meta/state` `env_vars`), regardless of the activation state of the process that built it.

### Suggested fix

In `_command()`, pass a sanitized `env=` to the `logged_run(activator_cmd, ...)` call that strips `CONDA_SHLVL`, `CONDA_PREFIX`, `CONDA_DEFAULT_ENV`, `CONDA_PROMPT_MODIFIER`, `_CE_M`, `_CE_CONDA` before invoking `conda shell.cmd.exe activate`, so the subprocess always computes activation "from scratch" instead of diffing against whatever state the builder process happens to be in.

### Environment

- conda: >= 25.3.0 (uses the `.env`-file activation path in `win.py`, not the older `.bat` path)
- OS: Windows

### Conda Info

```shell

```

### Conda Config

```shell

```

### Conda list

```shell

```

### Additional Context

_No response_

marcoesters:
### Description

Simplify the convoluted activation behavior on Windows by replacing the chained `cmd.exe`-PowerShell-`cmd.exe` calls and the manual activator parsing with a minimized `cmd.exe` instance that call `activate.bat`.

The current behavior chains a `cmd.exe` call that calls PowerShell with a hidden window, which then runs `start` to run the application inside an activated environment. This level of nesting may trigger Windows security software (see #468). The reason for this setup was that the flash to start PowerShell in a hidden window was shorter than starting the process from `cmd.exe` directly. Alternatives were explored in #468 by @lrandersson and @jacobwilliams, including setting the Window status of the shortcut.

Another issue was introduced due to the changed activation behavior in `conda` (see #321), which changed from shell-commands to an environment file format. Parsing the environment file format has a major design flag: it runs on shortcut creation, which makes it depend on the shell state at install time, not run time (see #545). The reason activators are used in `menuinst` is likely because in general, `conda` cannot activate an environment without being initialized. An exception, however, is `cmd.exe` by using `activate.bat`.

This PR simplifies the way activation happens in `conda`. It uses `activate.bat` instead of calling an activator, which activates the environment with the same behavior independent of the state at install time. The shortcuts then starts minimized, which flashes a console icon in the task bar. This is less intrusive than a flashing of a PowerShell window in the middle of the screen though. The application is then started in a detached process to close the console window immediately after activation.

Closes #321, closes #468, closes #545 

### Checklist - did you ...


menuinst/platforms/win.py
```diff
@@ -339,38 +334,15 @@ def _write_script(self, script_path: os.PathLike | None = None) -> Path:
...
-                    "START",
-                    "/MIN",
-                    '""',  # START requires this empty window title (must remain quoted)
-                    str(pwsh),
-                    "-WindowStyle",
-                    "hidden",
-                    start_script,
-                ]
-
+            cmd_exe = Path(os.environ.get("SystemRoot", "C:\\Windows")) / "system32" / "cmd.exe"
```
jaimergp, line 341: I wonder if we should try to respect `COMSPEC` and friends.

src/winshortcut.cpp
```diff
@@ -44,6 +44,8 @@ static PyObject *CreateShortcut(PyObject *self, PyObject *args)
     int iconindex = 0;
     PyObject *py_workdir = NULL;
     PyObject *py_app_id = NULL;
+    // https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-showwindow
```
jaimergp, line 47: Good find!

jaimergp:
In general, that new flag in the shortcut creation makes all the difference, so good find! As long as this method doesn't change the UX for users too much (that CMD flash was really annoying) and everything else works as intended, I'm ok with this PR. It's well documented too, thanks for that!

That aside, the `activate.bat` changes are orthogonal to the security-oriented improvement here, right? I'd rather have it in a separate PR, because I recall incidents where `activate.bat` was stale with respect to the activator logic (e.g. misplaced `conda` binary etc), and we may need to revert that one.

marcoesters:
> I recall incidents where activate.bat was stale with respect to the activator logic (e.g. misplaced conda binary etc),

Could you elaborate? Was the root cause in `conda`?

src/winshortcut.cpp
marcoesters: I can't take all the credit here - this was explored in #468 by @jacobwilliams and @lrandersson, I just integrated it here.

